PR #54, opened 2018-07-20
Plural date-time-format-error for "date and time"

comradekingu:
Alternatively, "due to not matching the required format", which works for all.

Is it specifically both failing, or just one or more?
Should I bump the version number and date of the maven.buildNumber.plugin properties file manually?


cpesch:
Dear Allan,

it's the combined input of date and time in one field that is failing.

Your pull request is perfect just the way it is. Please don't bump version numbers manually.

PR #69, opened 2022-05-19
Fixed ListCellRender problems with GTK look.

ds10git:
With GTK look especially JComboBox with DefaultListCellRenderer has some issues looking like intended for GTK look.
RouteConvert-GTK-Look-problems

After changes to use the ListCellRenderer already set for the JComboBox at creation as backend for formating the JLabel the GTK look is without those issues for JComboBox:
RouteConvert-GTK-Look-fixed

### DIFF
--- a/common-gui/src/main/java/slash/navigation/gui/renderer/BackendListCellRenderer.java
+++ b/common-gui/src/main/java/slash/navigation/gui/renderer/BackendListCellRenderer.java
@@ -1,0 +1,29 @@
+package slash.navigation.gui.renderer;
+
+import java.awt.Component;
+
+import javax.swing.DefaultListCellRenderer;
+import javax.swing.JLabel;
+import javax.swing.JList;
+import javax.swing.ListCellRenderer;
+
+public class BackendListCellRenderer extends DefaultListCellRenderer {
+	private ListCellRenderer backend;
+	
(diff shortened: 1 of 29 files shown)


cpesch:
I understand the problem, thank you very much for the screenshot.

I don't agree with the solution:

1.  It can't be the task of each combobox renderer to include a workaround for a look and feel.
2. I don't understand why the workaround works: why does calling the default combobox renderer (except for JLabel) have a different effect than instantiating a DefaultListCellRenderer?

ds10git:
This is simply the case for the GTK look, the renderer for the combobox for GTK seems to work different than for any other look and feel. It adds padding with borders and handles disabling somehow different. And this problem is persistent like forever. The only other way I know to fix this would be to scrap the renderer classes for the JComboBoxes and rely on the toString() method of the components displayed in the list. Using renderer classes derived of DefaultListCellRenderer will always result in destroyed layout with GTK. Without this workaround or using toString() RC will not be displayed correctly with GTK.

cpesch:
I'm thinking about centralizing the GTK workaround in a base renderer class to avoid spreading it upon the code base. I don't want to harm the 98% non-Linux users with it. And passing the current renderer of each combobox instance to the renderer is really error prone.

Have to setup a Linux VM first to test.

PR #71, opened 2022-05-23
Kml name fix

ds10git:
When loading KML files that were saved by RouteConverter the name of the position lists keep adding "/Waypoints:" or "/Track:" at the beginning. So this change checks for those and removes them.

On small screens the popup of the JComboBoxes for map and map style selection are too small to be able to see the full description of the entries, so one sometimes doesn't see what to select. This changes the popup size to be at least as wide as the entry with the widest description.

### DIFF
--- a/common-gui/src/main/java/slash/navigation/gui/renderer/BackendListCellRenderer.java
+++ b/common-gui/src/main/java/slash/navigation/gui/renderer/BackendListCellRenderer.java
@@ -1,0 +1,29 @@
+package slash.navigation.gui.renderer;
+
+import java.awt.Component;
+
+import javax.swing.DefaultListCellRenderer;
+import javax.swing.JLabel;
+import javax.swing.JList;
+import javax.swing.ListCellRenderer;
+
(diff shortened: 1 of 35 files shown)


cpesch:
That's actually 3 changes in 1 – please separate them to make review easier. 

[Fixed ListCellRender problems with GTK look.](https://github.com/cpesch/RouteConverter/pull/71/commits/c7d65cfbe8b8c77102ee8e0a9c668283bf9cc27e)
I'm not convinced about this one as I've pointed out in the earlier PR.

[If existing remove leading "/Waypoints:", "/Tracks:" from name of route](https://github.com/cpesch/RouteConverter/pull/71/commits/b72d67f2749839e6e7e79a3bc4859538cb6d99bc) 
Sounds interesting, I didn't have time yet to extract it and see the effect.

[Make popup of map JComboBoxes at least as wide](https://github.com/cpesch/RouteConverter/pull/71/commits/8968aa8e73c61cd0dceb931016f8e74cb884a9ad) 
There seems to be code missing – I cannot see the ItemWideSomething base class

PR #72, opened 2022-05-28
Make popup of map JComboBoxes at least as wide.

ds10git:
I've separated this so it should be easier to review now.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapSelector.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapSelector.java
@@ -23,6 +23,7 @@ import com.intellij.uiDesigner.core.GridConstraints;
 import com.intellij.uiDesigner.core.GridLayoutManager;
 import org.mapsforge.map.model.IMapViewPosition;
 import slash.navigation.gui.Application;
+import slash.navigation.maps.item.ItemWideComboBox;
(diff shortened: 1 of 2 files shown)


cpesch:
Again a GTK problem? I don't see problems with MacOS and Windows.

ds10git:
No this is not a GTK problem. It's a problem for smaller screen resolutions. My notebook has a screen resolution of 1366x768 pixels, so the checkboxes can be only as wide as the screen allows and the standard popup window of checkboxes is only as wide as the checkbox itself, even if the screen has a lot more room. This is independent of the look and feel and will also occur on Windows for example.

Here a part of a screenshot that uses the Nimbus look and feel and the window is maximized so there is no more room for the checkbox:
default-combobox

With the wide combobox it looks like this:
widecombobox

In the first screenshot one cannot see if the map is for Sachsen or Sachsen-Anhalt. When using GTK look and feel the font size is bigger so even less text is visible.

cpesch:
That's the default behaviour on a Mac:

Bildschirmfoto 2022-07-01 um 17 44 45

And that's on Windows:

Bildschirmfoto 2022-07-01 um 17 54 33

So I guess this is LookAndFeel specific.

cpesch:
That's the default behaviour on a Mac:

Bildschirmfoto 2022-07-01 um 17 44 45

And that's on Windows:

Bildschirmfoto 2022-07-01 um 17 54 33

So I guess this is LookAndFeel specific.

PR #73, opened 2022-05-28
If existing remove leading "/Waypoints:", "/Tracks:" from name of route

ds10git:
I've separated this so it should be easier to review now.


cpesch:
Solved with https://github.com/cpesch/RouteConverter/commit/2b87954234582e6fadfb0d9b5e5c67cd80ad8d89

PR #106, opened 2026-06-18
fix(download): drop ETag on outdated file so re-download is unconditional

cpesch:
## Problem

A corrupt or outdated local target whose checksum no longer matches kept its stored ETag. `scanForOutdatedFilesInQueue` marked it `Outdated` and re-queued it, but the re-download still sent `If-None-Match`, so the server answered `304 Not Modified` and the bad file was never overwritten — an endless loop.

Observed in the wild (error report 1311): a stale `world.map` (1403310 bytes vs expected 111927, a 2014 file) survived every restart across 13 sessions / builds 74–89, leaving the map blank.

## Fix

Clear the ETag when marking a download `Outdated` so the re-download issues an unconditional GET, gets 200 with a body, and replaces the file.

## Test

Adds a hermetic regression test (`DownloadManager304Test`) covering the full corrupt → re-fetch cycle: corrupt the target, scan marks it Outdated + drops ETag, re-download serves a 200 body and restores the file. `download/` module green (75 tests).

### DIFF
--- a/download/src/main/java/slash/navigation/download/DownloadManager.java
+++ b/download/src/main/java/slash/navigation/download/DownloadManager.java
@@ -307,6 +307,10 @@ public class DownloadManager {
                 if (!validator.isChecksumsValid()) {
                     log.info("Found outdated download " + download);
 
+                    // drop the ETag so the re-download issues an unconditional GET:
+                    // a corrupt/outdated local file must be re-fetched, otherwise the
+                    // conditional request keeps returning 304 and the file is never replaced
+                    download.setETag(null);
                     download.setState(Outdated);
                     getModel().updateDownload(download);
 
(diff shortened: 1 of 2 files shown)


cpesch (commented):

Minimal, targeted fix for a real regression: a corrupt local file retained its ETag and looped on 304 responses across sessions. The ETag is cleared atomically with the `Outdated` state transition and persisted via `updateDownload`. The regression test covers the full corrupt→scan→re-fetch cycle with direct assertions on both state and the cleared ETag.

### download/src/test/java/slash/navigation/download/DownloadManager304Test.java

- **nit** [correctness] line 147 — The re-download leg creates a fresh `Download second` via `queueForDownload` rather than re-processing the original `download` instance that was marked `Outdated`. If `queueForDownload` always vends a new object with no prior ETag, `bodiesServed == 2` would pass regardless of the fix. The `assertNull(download.getETag())` assertion at line 139 guards the production change directly; this is a test-completeness gap, not a production correctness issue.

---

PR #108, opened 2026-06-18
test: raise coverage across 5 weak spots tied to recent bugs

cpesch:
Five focused test additions, each targeting a real recent failure or a thin-coverage hotspot. ~27 new tests; all hermetic (no network/fixtures), run in the gated unit phase.

**1. Comparator total-order contract** (`graphhopper`)
Adversarial fixed-seed sets (contained/overlapping/disjoint/identical/null bounding boxes) through `GraphDescriptorComparator` + the `PREFERRED` comparator; asserts antisymmetry, transitivity, and that `List.sort` (TimSort) never throws. Includes a self-test proving the harness catches a known-intransitive comparator. Guards the "Comparison method violates its general contract" class.

**2. Download state machine** (`download`)
Local-`HttpServer` tests: Extract with/without fragments, resume via HTTP 206, and conditional-GET 304 with a corrupt local target re-fetching unconditionally. Covers the ETag/304 cache-poison (#106) and gpsbabel Extract-fragments paths.

**3. Background-map failure modes** (`mapsforge-mapview`)
`setBackgroundMap` with a corrupt/missing `.map` must log + null the layer, not let the mapsforge header-check throw escape on the EDT (the blank-map-with-clean-log of #107).

**4. Format negative tests** (`navigation-formats`)
GPX and KML are the most-used formats but were covered mostly by happy-path roundtrips (and the broad roundtrip suite is an excluded `*IT`). Empty/truncated/non-format/malformed input → clean `IOException`; valid-but-empty → no routes; a tiny valid GPX 1.1 still parses lat/lon/ele — so a parse regression fails the gated build.

**5. Opt-in `STRICT_MODULES` gate** (`scripts/verify-runtime.sh`)
The jdeps module-superset check is advisory (jdeps over-approximates on a fat jar). `STRICT_MODULES=1` promotes it to a hard gate naming missing modules; default stays advisory (#3/#4 remain the real runtime gates).

All five module suites green locally.

### DIFF
--- a/download/src/test/java/slash/navigation/download/DownloadManagerStateMachineTest.java
+++ b/download/src/test/java/slash/navigation/download/DownloadManagerStateMachineTest.java
@@ -1,0 +1,262 @@
+/*
+    This file is part of RouteConverter.
+
+    RouteConverter is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    RouteConverter is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with RouteConverter; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+    Copyright (C) 2007 Christian Pesch. All Rights Reserved.
+*/
+package slash.navigation.download;
+
+import com.sun.net.httpserver.HttpServer;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.util.EnumSet;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+import static java.io.File.createTempFile;
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static slash.common.io.Files.generateChecksum;
+import static slash.common.io.InputOutput.readFileToString;
+import static slash.navigation.download.Action.Copy;
+import static slash.navigation.download.Action.Extract;
+import static slash.navigation.download.State.Outdated;
+import static slash.navigation.download.State.Succeeded;
+
+/**
+ * Hermetic (no external network) tests of the download state machine, covering the paths that
+ * recently broke: ETag/304 conditional GET with a corrupt local target, Extract with/without
+ * fragments, and resume/partial-content. Uses a local {@link HttpServer} so they run in the
+ * unit (Surefire) phase rather than the network-gated {@link DownloadManagerIT}.
+ *
+ * @author Christian Pesch
+ */
+public class DownloadManagerStateMachineTest {
+    private static final String ENTRY_NAME = "first/second/447bytes.txt";
+    private static final String ENTRY_BODY = "Lorem ipsum dolor sit amet, consectetur adipisicing elit";
+    private static final String PLAIN_BODY = "0123456789ABCDEFGHIJ0123456789ABCDEFGHIJ"; // 40 bytes
+    private static final String ETAG = "\"state-machine\"";
+
+    private HttpServer server;
+    private final AtomicInteger bodiesServed = new AtomicInteger();
+    private final AtomicInteger partialServed = new AtomicInteger();
+    private final AtomicInteger conditional304 = new AtomicInteger();
+    private byte[] zipBytes;
+
+    private DownloadManager manager;
+    private File queueFile, target, targetDirectory;
+
+    @Before
+    public void setUp() throws IOException {
+        zipBytes = buildZip(ENTRY_NAME, ENTRY_BODY.getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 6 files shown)


cpesch (commented):

Pure test addition across five modules plus a shell script opt-in gate, with no production code changes. All six findings are minor or nit-level: two null-guard weaknesses that would silently pass on a null return, one missing pre-condition in a mapsforge test, an over-latitude random-box generator, and two nits on internal-API coupling and reflection-brittle field access.

### download/src/test/java/slash/navigation/download/DownloadManagerStateMachineTest.java

- **nit** [convention] line 23 — `com.sun.net.httpserver.HttpServer` lives in the `jdk.httpserver` module, an internal JDK API surface. Works fine with standard Surefire + JDK 17, but future module-path tightening would break these tests at startup rather than at compile time.

### graphhopper/src/test/java/slash/navigation/graphhopper/GraphDescriptorComparatorContractTest.java

- **nit** [correctness] line 156 — `randomBox` generates `maxLat` up to `(minLat=59) + (height=60) = 119°`, exceeding the geographic 90° ceiling. If `BoundingBox` validates latitude bounds this will throw unexpectedly; if it does not, the randomised set silently exercises out-of-domain values the comparator may not be designed to handle.
- **nit** [convention] line 119 — `getDeclaredField("PREFERRED_GRAPH_DESCRIPTOR_COMPARATOR")` couples the test to a private field name. A rename of that field silently breaks the test at runtime rather than at compile time.

### mapsforge-mapview/src/test/java/slash/navigation/mapview/mapsforge/MapsforgeMapViewTest.java

- **minor** [correctness] line 202 — `assertNull(backgroundLayerOf(mapView))` is trivially true if `@Before` never sets `backgroundLayer` non-null. The test does not prove the error-handling path actively clears the field; a no-op `setBackgroundMap` would also pass. Consider calling `setBackgroundMap` with a valid file first to establish a non-null baseline, then verify the field transitions to null after the corrupt-file call.

### navigation-formats/src/test/java/slash/navigation/gpx/GpxFormatNegativeTest.java

- **minor** [correctness] line 105 — `if (routes != null)` guard silently passes when the parser returns null. A regression returning null instead of an empty list would go undetected. Use `assertNotNull` first.
```suggestion
        assertNotNull(routes);
        for (GpxRoute route : routes)
            assertEquals(0, route.getPositionCount());
```

### navigation-formats/src/test/java/slash/navigation/kml/KmlFormatNegativeTest.java

- **minor** [correctness] line 76 — Same `if (routes != null)` issue as `GpxFormatNegativeTest`: a null return from the KML parser passes the test without any assertion.
```suggestion
        assertNotNull(routes);
        for (KmlRoute route : routes)
            assertEquals(0, route.getPositionCount());
```

---

cpesch:
🤖 Patcher pushed a fix to `test/coverage-batch`. Re-running CI now.

cpesch (commented):
## Verdict: comment-only · confidence 0.93

Pure test addition (5 files) targeting documented recent bugs, plus an additive opt-in gate in `scripts/verify-runtime.sh`. No production code changed. All findings are minor or nit; no blockers.

### download/src/test/java/slash/navigation/download/DownloadManagerStateMachineTest.java

- **minor** [correctness] line 165 — All four tests call `manager.waitForCompletion()` with no test-level timeout. If `DownloadManager.waitForCompletion()` has no internal bound, a stalled download or a busy CI runner losing a loopback packet will hang the entire Surefire phase indefinitely.

Verify that `DownloadManager.waitForCompletion()` has a built-in timeout; if not, consider adding a JUnit `@Rule Timeout` or wrapping each call with a `CountDownLatch` and `await(30, SECONDS)` to bound test runtime.

### navigation-formats/src/test/java/slash/navigation/kml/KmlFormatNegativeTest.java

- **minor** [correctness] line 76 — `testValidButEmptyKmlReturnsNoRoutesWithoutThrowing` guards the loop with `if (routes != null)` so a null return passes the test silently — no assertion fires. The GPX equivalent uses `assertNotNull(routes)` before the loop.

```suggestion
        assertNotNull(routes);
        for (KmlRoute route : routes)
            assertEquals(0, route.getPositionCount());
```

- **nit** [convention] line 47 — `KmlFormatNegativeTest.read()` uses `new ParserContextImpl<>()` (no-arg) while `GpxFormatNegativeTest.read()` uses `new ParserContextImpl<>(null, null)`. If both forms are equivalent, align them for consistency.

---

cpesch:
🤖 Patcher pushed a fix to `test/coverage-batch`. Re-running CI now.

cpesch (commented):

Five focused hermetic test additions targeting real recent failures (comparator contract, download state machine, mapsforge failure modes, GPX/KML negative parsing) plus an opt-in `STRICT_MODULES` gate in `verify-runtime.sh`. Structure and coverage rationale are solid; one minor assertion gap in the KML test and three nits below.

### navigation-formats/src/test/java/slash/navigation/kml/KmlFormatNegativeTest.java

- **minor** [correctness] line 68 — `testValidButEmptyKmlReturnsNoRoutesWithoutThrowing` silently passes with zero assertions when the parser returns null; the `if (routes != null)` guard causes the loop body to be skipped entirely. The equivalent GPX test uses `assertNotNull`. If null is a documented valid return for empty KML a comment should say so; otherwise harden the assertion (also requires adding `import static org.junit.Assert.assertNotNull;`).

```suggestion
        List<KmlRoute> routes = read(
                "<?xml version=\"1.0\"?><kml xmlns=\"http://www.opengis.net/kml/2.2\"><Document/></kml>");
        assertNotNull(routes);
        for (KmlRoute route : routes)
            assertEquals(0, route.getPositionCount());
```

### scripts/verify-runtime.sh

- **nit** [correctness] line 91 — trailing `\"` inside the double-quoted `echo` string appends a literal `"` to the warning message output; the original warning line did not have this character.

```suggestion
    echo "::warning::fat jar references module(s) not shipped: ${extras[*]} (dead refs unless #4/#3 fail; set STRICT_MODULES=1 to gate)"
```

### download/src/test/java/slash/navigation/download/DownloadManagerStateMachineTest.java

- **nit** [style] line 27 — `com.sun.net.httpserver.HttpServer` is an internal JDK API. It is a pragmatic and reliable choice for a lightweight test HTTP server on the Java 17 baseline, but a brief comment noting the deliberate use of the internal package would help if `module-info` or `jlink` stripping is ever tightened.

### graphhopper/src/test/java/slash/navigation/graphhopper/GraphDescriptorComparatorContractTest.java

- **nit** [style] line 128 — `field.setAccessible(true)` on `PREFERRED_GRAPH_DESCRIPTOR_COMPARATOR` works within the same named module on Java 17 but will throw `InaccessibleObjectException` if an explicit `opens` restriction is added to a future `module-info` for `graphhopper`. Low risk today; worth a note if module descriptors are ever introduced.

---

PR #112, opened 2026-06-21
ci: SignPath code signing on prerelease + release (test policy)

cpesch:
Implements rc-meta specs/00043 (Windows code signing via SignPath Foundation), `gh-actions-integration` phase.

## What

- **`_build-windows.yml`** — Authenticode-sign the merged installers (`RouteConverterWindows.exe`, `TimeAlbumProWindows.exe`, `RouteConverterPortable.paf.exe`) via the SignPath `windows-installers` artifact config. Behind a `sign` input (default `false` → PR/fork path byte-for-byte unchanged) + `signing-policy-slug` (default `test-signing`).
- **`_build-linux-mac.yml`** — `jar-sign` the standalone downloadable jars (`RouteConverterLinux.jar`, `RouteConverterCmdLine.jar`) via the `jars` config; submitted in isolation so the config has no unlisted `.tar.gz`/`.app.zip`.
- **Portable rename moved into the build job** so signing uses exact paths (no wildcard); both `prerelease.yml` and `release.yml` drop their now-redundant version-suffix step. `previous-releases/` archive now carries the short name.
- **`prerelease.yml` + `release.yml`** — `sign: true` + `secrets: inherit` (both on `test-signing`).
- Folds the post-00044 verify-cut cleanup; strips stale comments from release/prerelease.

## Credential model

Cert stays in the SignPath HSM. CI holds only the submitter token `SIGNPATH_API_TOKEN` (cannot extract the cert) + `SIGNPATH_ORG_ID` var.

## Before this signs live (maintainer, SignPath UI + repo)

- [x] Create artifact configs `windows-installers` (zip → authenticode) + `jars` (zip → jar-sign) — project slug `RouteConverter`
- [x] `gh secret set SIGNPATH_API_TOKEN` + `gh variable set SIGNPATH_ORG_ID`

## ⚠️ Note

`release.yml` signs with the **test** cert until the production OV cert is issued (SignPath `release-signing` policy currently INVALID / CSR PENDING). Released installers won't pass SmartScreen until `release-sign-live` flips `signing-policy-slug` → `release-signing`.

https://claude.ai/code/session_01PxtaZK4t4XiD6GAufeabAf


cpesch (commented):

Focused CI addition wiring SignPath HSM signing for Windows installers and standalone Linux jars behind a `sign: false` default so PR/fork builds are byte-for-byte unchanged. Cleanup of `prerelease.yml`, `release.yml`, and `verify-cut.yml` is consistent with the portable-rename move into the build job and the post-00044 Mac-jar removal. Credential model (token in GitHub secrets, cert in SignPath HSM) is sound and the signing steps are correctly gated. No convention-check violations.

### .github/workflows/_build-linux-mac.yml

- **minor** [correctness] line 116 — `TimeAlbumProLinux.jar` is collected into `artefacts/` but is not submitted to SignPath for signing; only `RouteConverterLinux.jar` and `RouteConverterCmdLine.jar` appear in the `to-sign/` bundle and the overlay step. The Windows job signs `TimeAlbumProWindows.exe`, so the Linux counterpart is asymmetric. If the `jars` artifact config intentionally excludes TimeAlbumPro Linux (different policy, different phase, or commercial-signing path), a brief inline comment in the step body would prevent future confusion.

### .github/workflows/_build-windows.yml

- **nit** [correctness] line 190 — `cp RouteConverterPortable/target/classes/RouteConverterPortable*.paf.exe artefacts/RouteConverterPortable.paf.exe` will fail with "not a directory" if the glob ever matches more than one file; the original for-loop silently last-wins in that case. Low probability given the current build, but the `cp`-to-a-filename form is more fragile than the previous approach if the build ever changes to produce multiple `.paf.exe` outputs.

---

cpesch (commented):
## Verdict: comment-only · confidence 0.91

CI-only PR wiring SignPath Authenticode signing (Windows installers) and jar-signing (Linux/Mac standalone jars) behind a `sign` input flag defaulting to `false`, keeping PR/fork builds byte-for-byte unchanged. Portable-filename normalisation is folded into the build job, Mac `.jar` build intermediates are dropped from published artefacts, and `verify-cut.yml` is updated to the resulting 8-asset set. Two minor observations noted; no blockers.

### .github/workflows/verify-cut.yml

- **minor** [convention] line 77 — Removes the 301-redirect smoke-test for the two retired Windows exe names (`RouteConverterWindowsBundle.exe`, `RouteConverterWindowsOpenSource.exe`). specs/00044 §D mandates those server-side redirects permanently; dropping the verification means a regression in the redirect layer would go undetected by `verify-cut`.

### .github/workflows/release.yml

- **nit** [docs] line 20 — `signing-policy-slug` is not passed explicitly to `build-linux-mac` or `build-windows`, so the default `test-signing` applies to production release builds. The PR body documents the follow-up (`release-sign-live`), but an inline comment marking the TODO would prevent the production flip from being overlooked once the OV cert is issued.

---

cpesch (commented):

CI-only PR wiring SignPath code signing into prerelease and release workflows via opt-in `sign` inputs that default to `false`, leaving PR/fork builds byte-for-byte unchanged. The Portable rename moved into the build job is a clean defensive simplification; `verify-cut.yml` correctly tracks the 10→8 asset-count drift from removing Mac `.jar` intermediates. Two minor points below; no convention-check violations.

### `.github/workflows/_build-linux-mac.yml`

- **minor** [correctness] line 113 — `TimeAlbumProLinux.jar` is copied into `artefacts/` and is a standalone downloadable (it remains in `verify-cut.yml`'s expected asset list), but is absent from both the `to-sign/` bundle and the overlay step. If the SignPath `jars` artifact config intentionally excludes TAP Linux jar, a brief comment would clarify; if it should be signed, the Collect and Overlay steps both need it.

### `.github/workflows/release.yml`

- **nit** [docs] line 22 — `signing-policy-slug` is not passed explicitly, so both build jobs inherit the `test-signing` default for production releases. The PR body explains this is intentional until the OV cert is issued, but there is no in-workflow marker; a TODO comment at the `sign: true` line would make the future `release-sign-live` cut-over discoverable without consulting the PR description.

```suggestion
      sign: true
      signing-policy-slug: test-signing  # TODO: flip to release-signing once prod OV cert is issued (specs/00043)
```

---

cpesch (commented):

Well-structured gate-guarded SignPath integration: signing defaults to off and is enabled only for prerelease/release runs via explicit `sign: true`. Portable rename consolidation (moved into the build job so signing uses exact paths) and stale-comment cleanup are clean. One nit: the `test-signing` policy has no in-workflow durability marker to prompt the `release-signing` flip once the OV cert is issued.

### .github/workflows/release.yml

- **nit** [docs] line 23 — `signing-policy-slug` is not set in the `with:` block, so it silently inherits the `test-signing` default from the reusable workflow. The PR ⚠️ note documents the planned flip to `release-signing`, but workflow files outlast PR descriptions; an explicit inline comment would make the intent durable.
```suggestion
      sign: true
      signing-policy-slug: test-signing  # TODO: flip to release-signing once OV cert CSR resolves (specs/00043)
```

---

cpesch:

**Reason:** empty log_tail: no diagnostic output available — cannot determine which job failed or what the root cause is
**Confidence:** 0.00

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-3676` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-3676).

cpesch (commented):

Clean SignPath code-signing integration: signing is gated behind a `sign` boolean (default false, so PRs and forks remain byte-for-byte unchanged), the cert stays in the SignPath HSM, and the artifact-name patterns added to the publish jobs correctly exclude the unsigned interim artifacts from the final download step. One nit: the SignPath action is pinned to a mutable major-version tag while the same PR pins the Codecov action to a patch version.

### .github/workflows/_build-linux-mac.yml

- **nit** [security] line 124 — `signpath/github-action-submit-signing-request@v2` uses a mutable major-version tag; this PR already pins `codecov/codecov-action` to `@v5.5.5` — resolve `@v2` to a full commit SHA for equivalent supply-chain protection.

### .github/workflows/_build-windows.yml

- **nit** [security] line 202 — Same: `signpath/github-action-submit-signing-request@v2` should be pinned to a commit SHA rather than the mutable `@v2` tag.

---

cpesch:

**Reason:** Cannot resolve signpath/github-action-submit-signing-request@v2 to a commit SHA without network access; guessing a SHA is unacceptable for a supply-chain security fix
**Confidence:** 0.00

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-3678` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-3678).

cpesch (commented):

Adds SignPath code-signing integration for Windows installers and standalone JARs behind a `sign` boolean input (default false, so PR/fork paths are byte-for-byte unchanged), removes Mac intermediate JARs from published artefacts, and bumps the Maven wrapper to 3.3.4/Maven 3.9.9. The implementation is clean, the signing guard logic is correct, and the verify-cut asset count (10 → 8) is updated consistently. One minor security hygiene point on the signing action pin.

### .github/workflows/_build-linux-mac.yml

- **minor** [security] line 130 — `signpath/github-action-submit-signing-request@v2` uses a mutable floating tag. A tag can be silently moved to a different commit, so a supply-chain event could swap the action that processes release JARs. For a signing step that gates published artefacts, pinning to the full upstream commit SHA is standard hardening.
```suggestion
      uses: signpath/github-action-submit-signing-request@<UPSTREAM-COMMIT-SHA>  # v2
```

### .github/workflows/_build-windows.yml

- **minor** [security] line 208 — Same mutable `@v2` floating tag on `signpath/github-action-submit-signing-request` as in `_build-linux-mac.yml`. Pin to a full commit SHA for the same reason.
```suggestion
      uses: signpath/github-action-submit-signing-request@<UPSTREAM-COMMIT-SHA>  # v2
```

---

cpesch:

**Reason:** Cannot resolve upstream commit SHA for signpath/github-action-submit-signing-request@v2 — no network lookup available in this context; maintainer should obtain the SHA (e.g. `gh api repos/signpath/github-action-submit-signing-request/git/ref/tags/v2 --jq .object.sha`) and refire patcher with it, or apply the two-line change directly
**Confidence:** 0.00

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-3685` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-3685).

cpesch (commented):

SignPath integration is well-structured: `sign=false` default protects PR/fork builds, the portable-rename move into the build step correctly unblocks exact-filename signing, and the signature-presence gates (PowerShell `Get-AuthenticodeSignature` + `jarsigner`) prevent silent unsigned-file slip-through. The `pattern:` additions to `download-artifact` steps are a necessary safety measure that excludes the unsigned-intermediate artefacts from the publish jobs. Two minor security nits and one correctness nit; no blockers.

### .github/workflows/_build-linux-mac.yml

- **minor** [security] line 132 — Third-party action `signpath/github-action-submit-signing-request` is pinned by mutable semver tag `@v2`; if the tag is later repointed at malicious code the step has read access to `SIGNPATH_API_TOKEN` and all build artefacts.
```suggestion
          uses: signpath/github-action-submit-signing-request@<commit-sha>  # v2
```
- **nit** [correctness] line 156 — `grep -q '^jar verified\.'` leaves the trailing `.` unescaped (matches any character). Functionally harmless since `jarsigner` always emits a literal period, but `grep -qF` makes the intent explicit.
```suggestion
            if jarsigner -verify "artefacts/$j" | grep -qF 'jar verified.'; then
```

### .github/workflows/_build-windows.yml

- **minor** [security] line 209 — Same mutable-tag risk: `signpath/github-action-submit-signing-request@v2` without a commit-SHA pin.
```suggestion
          uses: signpath/github-action-submit-signing-request@<commit-sha>  # v2
```

---

Issue #101, opened 2026-06-01
Be able to show all routes and waypoints in the same file


Baltasarq:
Currently, when you have more than one list of points in RouteConverter, the edition of one points list hides all other routes or waypoints.

The proposal is basically to keep RouteConverter work as it does now, but draw all other list points, but grayed out. In summary: whatever points list you're editing (and you are only editing one at once), the other lists (waypoints, routes, whatever), are shown in gray or whichever color that makes it clear that they are not editable at this time.

Say you have the following configuration: you have a GPX with 1 route and 1 waypoint list. I start with only the route. I correct it, removing wrong turns or whatever. I'm still using RouteConverter the same way I've been using it. Then I create the waypoint list, and edit it. The route is still shown, but in gray. I can use RouteConverter the same way it's used today, everything is normal, apart from the fact that the other list, the route, is still displayed on the map, in gray, and I cannot edit it. I can only add waypoints. If I want to modify the route, then I would have to change back to the route in the combobox. Now I'm modifying the route, and the waypoints are still visible as gray markers.

cpesch:
Long thread in the forum: https://forum.routeconverter.com/thread-1447.html

Baltasarq:
I don't know any German, but I like very much the mockup from Mogh on the last page. My vision was not so ambitious.

cpesch:
Implemented — and it matches the less-ambitious vision you described, not the full panel.

Now every other position list in the file stays drawn on the map while you edit one:

- non-selected **routes/tracks** appear as a lighter shade of the active route/track color, so the darker, fully-saturated list you're editing clearly stands out
- non-selected **waypoint** lists show as subdued gray markers

They're display-only — you still edit one list at a time via the existing combo box, exactly as today; switch the combo and the emphasis follows. So with a GPX that has 1 route + 1 waypoint list you'll see both at once, with only the selected one editable.

(Mogh's richer forum design — per-list colors, visibility checkboxes, multi-edit — is noted as a possible future step, but this ships the read-only view you asked for.)

It's in the current **prerelease**: https://www.routeconverter.com/releases/

Could you try it with such a file and let me know whether the de-emphasis reads clearly on the map themes you use? Feedback on the styling/visibility is especially welcome.

Baltasarq:
Tested it! Works like a charm!
Yep, I have no problem in seeing the routes. I don't use any particular scheme.

cpesch:
[...]

PR #117, opened 2026-07-04
feat(mapview): show non-selected position lists read-only (gray)

cpesch:
Implements spec `00015` — issue #101 (gray read-only MVP).

## What
Every position list in a loaded file that is **not** the combo-selected one is drawn **gray, semi-transparent and non-interactive**. The selected list stays editable/colored exactly as today. Switching the combo makes the newly-selected list editable and the previously-selected one gray; both stay visible.

## Why the extra interface (`PositionListsModel`)
`MapView` lives in module `mapview`; `FormatAndRoutesModel` lives in `route-converter-gui`, which **depends on** `mapview`. Passing `FormatAndRoutesModel` into `MapView.initialize` directly would be a circular module dependency. A minimal `PositionListsModel` (`getRoutes` / `getSelectedRoute` + listeners) in `mapview`, which `FormatAndRoutesModel` extends, breaks the cycle with no caller changes.

## How
- `MapView.initialize` / `MapsforgeMapView.initialize` take the extra model; `RouteConverter.setMapView` passes it.
- Gray lists render into a dedicated `GroupLayer` inserted below the active list + selection markers, rebuilt from *routes − selected* on load, selection switch, list add/remove and color/line-width changes.
- Per characteristic: Waypoints → gray markers; Route/Track → one gray `Polyline` over stored positions. **No routing engine, no `DistanceAndTimeAggregator`** for the read-only set.
- Layer mutation is `synchronized` on the `GroupLayer` to exclude with the render thread's `draw()` (avoids a `ConcurrentModificationException` / blank map on switch-during-redraw).
- Read-only for free: plain `Marker`/`Polyline` have no drag/tap; all edit paths only consult the active `positionsModel`.
- Single-list files unchanged.

## Testing
- Full reactor build green (`route-converter -am`, Java 21).
- `mapview` + `mapsforge-mapview` unit tests pass (59).
- New `NonSelectedListsFlowIT` drives the model contract (routes-minus-selected + list events on load/switch) against a real multi-list GPX.
- High-effort multi-agent code review run; the one real bug it found (GroupLayer mutation race) is fixed in this branch.

## Still needs a human eyeball (native Swing GUI, not driveable headlessly here)
- gray lists actually render gray/semi-transparent below the colored active list;
- clicking/dragging never snaps onto a gray list.

## Out of scope (spec: future evolution)
Per-list colors, a list panel, visibility toggles, multi-list editing, multiple elevation profiles.

> Note for the closed edition: `MapView.initialize` gained a parameter. If `browser-mapview` (`JavaFX8WebViewMapView`) is still shipped there, it needs the new 4-arg signature too or it will `AbstractMethodError` at runtime. Not reachable in the open-source build.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
@@ -337,6 +337,7 @@ public class MapsforgeMapView extends BaseMapView {
         trackRenderer = new TrackRenderer(this, preferencesModel.getTrackColorModel(),
                 preferencesModel.getTrackLineWidthModel(), GRAPHIC_FACTORY);
         nonSelectedPositionListsRenderer = new NonSelectedPositionListsRenderer(this, positionListsModel,
+                preferencesModel.getRouteColorModel(), preferencesModel.getTrackColorModel(),
                 preferencesModel.getRouteLineWidthModel(), preferencesModel.getTrackLineWidthModel(), GRAPHIC_FACTORY);
     }
 
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
@@ -71,6 +71,7 @@ import slash.navigation.mapview.mapsforge.overlays.OverlayManager;
 import slash.navigation.mapview.mapsforge.renderer.BorderPainter;
 import slash.navigation.mapview.mapsforge.renderer.MagnifierPainter;
 import slash.navigation.mapview.mapsforge.renderer.MapViewLayerOperations;
+import slash.navigation.mapview.mapsforge.renderer.NonSelectedPositionListsRenderer;
 import slash.navigation.mapview.mapsforge.renderer.RouteRenderer;
 import slash.navigation.mapview.mapsforge.renderer.TrackRenderer;
 import slash.navigation.mapview.mapsforge.tiles.DefaultTileLayerFactory;
@@ -147,10 +148,13 @@ public class MapsforgeMapView extends BaseMapView {
     private static final byte MAXIMUM_ZOOM_LEVEL = 22;
 
     private PositionsModel positionsModel;
+    private PositionListsModel positionListsModel;
     private MapPreferencesModel preferencesModel;
     private MapViewCallbackOpenSource mapViewCallback;
+    private NonSelectedPositionListsRenderer nonSelectedPositionListsRenderer;
 
     private final TableModelListener positionsModelListener = new PositionsModelListener();
+    private final ListDataListener positionListsModelListener = new PositionListsModelListener();
     private final ListDataListener characteristicsModelListener = new CharacteristicsModelListener();
     private final ChangeListener routingPreferencesListener = e -> {
         if (positionsModel.getRoute().getCharacteristics().equals(Route))
@@ -164,6 +168,8 @@ public class MapsforgeMapView extends BaseMapView {
             this.waypointIcon = null;
         }
         this.updateDecoupler.replaceRoute();
+        // line widths apply to the gray set, too
+        updateNonSelectedPositionLists();
     };
     private final ChangeListener displayedMapListener = e ->
             handleMapAndThemeUpdate(true, !isVisible(this.mapView.getModel().mapViewPosition.getCenter()));
@@ -200,9 +206,11 @@ public class MapsforgeMapView extends BaseMapView {
     // initialization
 
     public void initialize(PositionsModel positionsModel,
+                           PositionListsModel positionListsModel,
                            MapPreferencesModel preferencesModel,
                            MapViewCallback mapViewCallback) {
         this.positionsModel = positionsModel;
+        this.positionListsModel = positionListsModel;
         this.preferencesModel = preferencesModel;
         this.mapViewCallback = (MapViewCallbackOpenSource) mapViewCallback;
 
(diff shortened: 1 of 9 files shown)


cpesch (commented):

Clean, well-tested maintainer change that renders non-selected position lists gray/read-only via a new minimal `PositionListsModel` interface to avoid a circular module dependency; the GroupLayer mutation race is guarded and a contract IT is added. No blockers found; two minor/nit items only. GUI rendering still needs a human eyeball as the author notes.

### mapview/src/main/java/slash/navigation/mapview/MapView.java
- **minor** [correctness] line 38 — The `MapView.initialize` interface method gained a parameter; any implementation not in the open-source reactor (the closed-edition `JavaFX8WebViewMapView` noted in the PR body) must adopt the 4-arg signature or it will `AbstractMethodError` at runtime — track the closed-edition update alongside this merge.

### route-converter-gui/src/test/java/slash/navigation/converter/gui/models/NonSelectedListsFlowIT.java
- **nit** [convention] line 1 — New test file omits the GPL license header that the sibling new production file (`PositionListsModel.java`) carries; other source files in this tree include it.

---

cpesch:
🤖 Patcher pushed a fix to `feature/00015-nonselected-lists-readonly`.

cpesch (commented):
## Verdict: approve · confidence 0.86

Second review pass — the prior patch round addressed the one real bug (GroupLayer mutation race), now fixed by rebuilding the gray set under `synchronized(nonSelectedPositionListsLayer)` on the same monitor `GroupLayer.draw()/onDestroy()` hold. The `PositionListsModel` interface breaks the `mapview` ↔ `route-converter-gui` cycle without caller changes, the gray set is derived as routes-minus-selected and rebuilt on the right list/selection/line-width events, and `NonSelectedListsFlowIT` guards that contract. No blockers, majors, or convention issues remain.

The description's own caveat stands: the gray/semi-transparent render below the active list and the no-snap-onto-gray behaviour are native Swing and need a human eyeball — not driveable headlessly here.

---

cpesch (commented):

Second-pass review — the prior round's GroupLayer mutation race is fixed (EDT clear/repopulate now holds the same `synchronized(layer)` monitor the render thread's `draw()`/`onDestroy()` use), the `mapview`→`route-converter-gui` cycle is broken cleanly by the minimal `PositionListsModel` interface with no caller changes, and `NonSelectedListsFlowIT` guards the routes-minus-selected contract plus the load/switch list events. No blocker or major issues and all convention checks pass. Confirming merge-readiness; a human still merges.

---

cpesch:
Rebased onto current master (force-pushed). Master had since merged the MapsforgeMapView refactors #119 (painters / tile-factory / overlay-manager), #125 (lambdas, UpdateDecoupler, coverage gate) and #126 (MapViewCalculations), which conflicted heavily with the original branch.

Rather than resolve a large blob conflict against the old structure, I squashed the branch to a single commit and re-derived the gray feature on top of the new structure:
- `NonSelectedPositionListsRenderer` plugs into the current renderer/overlay layout;
- `MapsforgeMapView` re-wired against the refactored form (lambda listeners, the split `createWaypointIcon(color, opacity)`, the layer-manager insertion);
- the standalone pieces (`PositionListsModel`, the `MapView`/`FormatAndRoutesModel`/`RouteConverter` wiring, `NonSelectedListsFlowIT`, spec) carried over unchanged.

Now MERGEABLE. Verified: full reactor build green (Java 21), `NonSelectedListsFlowIT` green, mapsforge-mapview 87 tests green with the coverage gate met. Still needs the manual UI eyeball before merge (gray lists render gray/below the active list; clicking-dragging never snaps onto a gray list).

cpesch:

Clean, well-documented maintainer PR: the module cycle is broken with a minimal `PositionListsModel` interface, the gray non-selected render path is isolated in its own renderer, the `GroupLayer` mutation race is handled by sharing the render-thread monitor, and it ships unit + model-contract tests. Two non-blocking items below; z-order and actual gray rendering still need the manual UI pass the author already flagged.

### mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/NonSelectedPositionListsRenderer.java
- **minor** [correctness] line 170 — The shared gray waypoint Bitmap is cached in `getWaypointIcon()` and reused across every `layer.layers.clear()`/rebuild but is never destroyed; confirm mapsforge Marker ref-counting does not leak (or prematurely free) the bitmap across repeated `update()` calls.

### specs/00015-show-nonselected-lists-readonly.md
- **nit** [docs] line 145 — Implementation note says non-selected waypoints render at opacity 0.5, but the code uses `NON_SELECTED_OPACITY = "0.7"`; align the note with the constant.

---

cpesch:

Read-only gray rendering for non-selected position lists: clean module-cycle break via the new `PositionListsModel` interface, `GroupLayer` mutation correctly synchronized against the render thread, and solid test coverage (`lighten()` unit tests + model-contract IT). The `MapView.initialize` signature change is safe in the open-source build (sole implementor `MapsforgeMapView` is updated; reactor build reported green). Two minor doc/defensive nits only.

### specs/00015-show-nonselected-lists-readonly.md
- **minor** [docs] line 161 — The implementation note says non-selected waypoint icons use opacity 0.5, but the shipped `NonSelectedPositionListsRenderer` sets `NON_SELECTED_OPACITY = "0.7"`; update the note to the shipped value.

### mapsforge-mapview/.../MapsforgeMapView.java
- **nit** [correctness] line 669 — `handleNonSelectedPositionLists()` dereferences `nonSelectedPositionListsRenderer` without the null guard that `updateNonSelectedPositionLists()` uses; add the same guard for symmetry in case the map/theme update path can run before the renderer is built.

---

cpesch:
**On the red codecov checks:** all three build jobs pass (Java 21, Java 21 on Windows, Java 25). The failing checks are codecov/patch and codecov/project only.

The uncovered patch lines are the GUI rendering path — `NonSelectedPositionListsRenderer.update()/addPositionList()` (needs a live map: `asLatLong`, `getTileSize`, `createWaypointIcon`), the `MapsforgeMapView` wiring/listener, and the `RouteConverter.initialize` call site. None of it runs in headless unit tests. The testable logic is covered: `lighten()` (unit test) and the routes-minus-selected model contract (`NonSelectedListsFlowIT`); the module's JaCoCo gate passes.

This is the same headless-GUI limitation that leaves other map/panel PRs with low patch coverage. Merging over the advisory codecov gate.

PR #118, opened 2026-07-05
spec 00011: crash diagnostics + opt-in anonymous telemetry (client)

cpesch:
Implements spec 00011 client side (Phase 1 local diagnostics + Phase 2 opt-in sender). Server endpoint + Phase 3 alerting are a separate rc-site PR.

## Phase 1 — local diagnostics (always on)
- `DiagnosticReport` — allowlist-only payload (schema_version first; root-cause class + cause chain; stack; priority-signature + missing_class; app version; OS/Java via `Platform`; `bundled_jre` marker). No paths, names, or user documents.
- `CrashReportSpool` — capped (10) on-disk spool under `temp/crash-reports`.
- `CrashHandler` seam + `Application` default-handler hookup with a ThreadLocal re-entry guard. EDT-during-modal exceptions reach the default handler on Java 17+ (verified by `EventDispatchThreadExceptionTest`; no custom `EventQueue` needed).
- `CrashReporter` — assemble → spool → prefill `SendErrorReportDialog`; one dialog per session; `offerSpooledReports()` at startup. Dialog gains a prefill entry point + "Copy diagnostic report" button.
- Bundled-JRE marker `-Drouteconverter.bundledJre=true` in the RC Mac/Windows launchers.

## Phase 2 — opt-in anonymous telemetry (default OFF)
- `CrashReportSender` — async, fail-silent HMAC-SHA256 sender (`X-RouteConverter-Signature` over the raw JSON body) → `v1/crash-report/`. Shared secret is a **placeholder** (`CHANGEME-crash-report-hmac-secret`) overridable via `-Drouteconverter.crashReportSecret`; a live value is injected at build/deploy, never committed.
- `MultipartRequest.setBody` — send an exact raw JSON body the server can sign-verify.
- Consent question + persistent preference; opted-in path silently flushes the spool, opted-out path is the manual dialog.

## Tests
common 194 · rest 16 · common-gui 61 · feedback 3 · route-converter-gui 199 — all green. New: `DiagnosticReportTest`, `CrashReportSpoolTest`, `EventDispatchThreadExceptionTest`, `CrashReportSenderTest`, `CrashReporterTest`.

Reactor `*FormatIT` failures (navigation-formats samples/network) are pre-existing and unrelated.


cpesch:

Adds spec 00011 crash diagnostics + opt-in telemetry with solid test coverage. One privacy concern: the diagnostic payload embeds full stack traces and exception cause messages, which routinely carry absolute file paths (and therefore OS usernames / document names), conflicting with the consent dialog's "no files, names or personal data" promise. Convention checks pass.

### common/src/main/java/slash/common/log/DiagnosticReport.java
- **major** [security] line 72 — cause_chain (exception messages) and stack_trace are captured verbatim into the telemetry payload; exception messages commonly embed absolute file paths (e.g. a FileNotFoundException on an opened GPX), which reveal the OS username and document names. This contradicts the opt-in consent text "No files, routes, names or personal data are included". Either sanitize/allowlist message text before sending, or correct the consent copy to reflect that stack traces/messages may contain such data.

### route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/CrashReporter.java
- **nit** [correctness] line 118 — in the opted-out path `offerSpooledReports()` only reads/deletes `spool.newest()`; older spooled reports from prior sessions are never surfaced and linger until evicted by the spool cap, so pre-cap crashes are effectively dropped for opt-out users.

---

common/src/main/java/slash/common/log/DiagnosticReport.java
```diff
@@ -0,0 +1,199 @@
...
+import static slash.common.helpers.ExceptionHelper.printStackTrace;
+import static slash.common.system.Platform.getJava;
+import static slash.common.system.Platform.getPlatform;
+
+/**
+ * An allowlist-only diagnostic payload assembled from an uncaught {@link Throwable}
+ * plus the runtime environment. It is rendered as a stable JSON object so a crash
+ * arrives as a full root-cause chain and environment instead of a screenshot of a
+ * truncated dialog.
+ * <p>
```
cpesch, line 38: shouldn't the "old" logging be switched to JSON and put next to the crash logging?

route-converter-gui/src/main/java/slash/navigation/converter/gui/RouteConverter.java
```diff
@@ -221,12 +225,15 @@ protected String getProduct() {
 
     protected void startup() {
         startupStartMillis = System.currentTimeMillis();
+        setCrashHandler(crashReporter);
         initializeLogging();
```
cpesch: unify both paths – crash handler and logging?

RouteConverterMac/src/main/app-resources/MacOS/RouteConverterOpenSource
```diff
@@ -26,4 +26,5 @@ exec "$DIR/bin/java" \
   -Xdock:icon="$DIR/Resources/RouteConverter.icns" \
   --add-exports java.desktop/com.apple.eawt=ALL-UNNAMED \
   -Xmx1024m \
+  -Drouteconverter.bundledJre=true \
```
cpesch, line 29: why is that necessary?

cpesch:
Pushed `a5f134014` addressing the review:

**Privacy — stack traces leaking paths (Inspector major, and the same finding on rc-site #2).** `DiagnosticReport` now scrubs the known `user.home` and `user.name` to `<USER_HOME>`/`<USER>` in `cause_chain`, `stack_trace` and `missing_class` before anything is spooled or sent — deterministic literal replacement of two known values, not a heuristic path mask (which would fail open). Corrected the javadoc and the consent copy (en/de) to say stack traces are included and the user name is stripped from paths, so the promise matches reality.

**Spooled reports only surfacing newest (Inspector nit).** `offerSpooledReports()` now offers every spooled report, not just `newest()` — a report the user dismisses without sending previously blocked all older ones forever.

**"unify both paths — crash handler and logging" (your comment, RouteConverter.java:229).** Folded `setCrashHandler` into a single `initializeDiagnostics()` alongside logging bring-up; the handler is still registered first so an early init crash is captured.

Two I want your call on rather than change blind:

**"switch the old logging to JSON and put it next to the crash logging" (DiagnosticReport:38).** The text log (`RouteConverter.log`) and the JSON crash payload serve different consumers — a human reading a debug log vs. machine-parseable root-cause telemetry — and the manual report dialog renders the log as text. Converting all app logging to JSON is invasive and would change that dialog. My recommendation is to keep them separate; the spool already lives in the same temp dir as the log. Happy to colocate/rename or go further if you'd rather unify — say the word.

**"why is that necessary?" (launcher `-Drouteconverter.bundledJre=true`).** It's the only reliable way to tell "running our shipped stripped JRE" from "running the user's system Java" — the core signal that separates *our bundle is broken* from *this user's unusual Java*. The spec explicitly ruled out inferring it from `java.home`. Alternative is a marker file shipped in the bundle instead of the JVM flag; the flag is one line per launcher and simplest. Can switch to a marker file if you prefer.

Tests green: `common` 196 (incl. 2 new scrub tests), `common-gui` 61, `rest` 16, `feedback` 3, `route-converter-gui` 199.

cpesch:

Well-structured, well-tested maintainer PR implementing spec 00011 client-side crash diagnostics + opt-in telemetry. The allowlist-only payload, `user.home`/`user.name` scrubbing, opt-out-by-default consent, and the fail-silent HMAC sender are all sound and covered by new tests. No blockers; all convention checks are clean. One minor concurrency note below.

### route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/CrashReporter.java

- **minor** [correctness] line 150 — `flushSpooledReports()` has no guard against concurrent invocation: a crash-triggered `flushSpooledReportsAsync()` overlapping the startup `offerSpooledReports()` flush lets two daemon threads list and send the same spool file before either deletes it, so an opted-in report can be sent twice.

---

cpesch:

**Reason:** feedback is an automated Inspector review (verdict: comment-only, confidence 0.90, no blockers) — not a maintainer change request. The one minor note (unsynchronized flushSpooledReports) is informational, not an instruction to act on.
**Confidence:** 0.0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-7691` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-7691).

PR #121, opened 2026-07-05
Browse columns polish: View-menu submenu, header sorting, all locales (spec 00012)

cpesch:
Polish backlog for the Browse Length/Duration columns feature:

1. **View-menu columns submenu** — `RoutesTableHeaderMenu` now binds to both the header popup and a new `View → show-route-column` submenu, mirroring how `PositionsTableHeaderMenu` uses `initializeShowColumn(findMenu(...))`.
2. **Header-click sorting** — `TableRowSorter` with per-column comparators (the downloads-table idiom): Name/Creator case-insensitive, Length/Duration numeric on the metadata values with missing-last (matches the `–` renderer rule); `setSortsOnUpdates(true)` keeps metadata fills correct. Added view→model index conversions in `openRoute`, drag handler and `RouteModelHelper` so selection/delete/rename/drag act on the right rows when sorted.
3. **Locales** — Length/Duration column keys + submenu key translated in 23 bundles by copying each bundle's own established `length=`/`duration=` values (no machine translation); stub bundles (ar, fi, lv, mr, tr) fall back to en.

Tests: 9 new comparator/sorter tests incl. headless `JTable` integration; route-converter-gui 208 tests green.

Reviewer note: the new submenu reuses the exact "Show Column" label of the positions-table submenu in every locale (deliberate — no invented translations). The two View submenus are distinguished by position only; if that reads confusingly, distinct labels need a proper translation pass.


cpesch:
## Verdict: request-changes · confidence 0.88

The PR is scoped (title, body, branch `spec/00012-polish`) to the Browse Length/Duration column polish, but the diff also ships an entire unrelated crash-telemetry subsystem (spec 00011 Phase 2). That accidental scope bleed is the blocking issue; the browse-column changes themselves (view↔model index conversions, comparators, `TableRowSorter`, locale keys) look correct.

### route-converter-gui/.../helpers/CrashReporter.java
- **major** [correctness] line 1 — This PR is described and branched as spec 00012 browse-column polish, but bundles a whole separate crash-telemetry feature (CrashHandler/CrashReporter/DiagnosticReport/CrashReportSpool/CrashReportSender, MultipartRequest raw-body, bundledJre launch flags, crash-consent prompt). These look like commits accidentally carried from another branch and should be split into their own PR so each is reviewed against its own spec.

### feedback/.../domain/CrashReportSender.java
- **minor** [security] line 48 — When the crash-report secret system property is unset, `sign()` falls back to the committed placeholder key, so the signature is computed with a publicly known key and carries no real authenticity. Consider skipping the send (or logging a config error) when the secret is still the default.

### route-converter-gui/.../helpers/FrameMenu.java
- **minor** [docs] line 111 — The new View → show-route-column submenu reuses the exact "Show Column" label of the positions-table submenu in every locale, so the two View submenus are indistinguishable except by position (you already flagged this). A distinct key/label would remove the ambiguity.

### rest/.../MultipartRequest.java
- **nit** [style] line 91 — The added `setHeader` override only delegates to `super` with no added behaviour; add a comment if it exists solely to widen visibility, otherwise it is redundant.

### route-converter-gui/.../panels/BrowsePanel.java
- **nit** [style] line 100 — The new `RoutesTableModel.*` static imports are interleaved between the two `JTableHelper` static imports, breaking the otherwise grouped import order.

---

cpesch:

**Reason:** Major finding asks to split an entire unrelated crash-telemetry feature (~15 files across common, common-gui, feedback, rest, route-converter-gui) out of this spec-00012 browse-column-polish PR. Actioning that means reverting a large fraction of the existing change set, which exceeds the 10-file/300-line patch cap and is equivalent to reverting the original commit — a scope/judgment call for the maintainer, not a mechanical one-iteration patch. Shipping only the minor/nit fixes (secret fallback, submenu label, redundant setHeader, import order) while leaving the blocking major finding unaddressed would be a partial fix that ignores the dominant part of the request, so no partial patch is applied either.
**Confidence:** 0.00

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-7680` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-7680).

cpesch (commented):

Maintainer-authored GUI polish: the view→model row-index conversions for the new `TableRowSorter` are correct across `openRoute`, the drag handler and `RouteModelHelper`, and the comparator logic is well covered by the new tests. Only minor/nit items remain.

### route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/FrameMenu.java
- **minor** [style] line 111 — The new `show-route-column` submenu sits in the same View menu as `show-column` and, across every locale, reuses both the identical label and the identical mnemonic (e.g. en 'C', hu 'O'), so the accelerator only ever reaches the first submenu — the author's own reviewer note flags this ambiguity.

### route-converter-gui/src/main/java/slash/navigation/converter/gui/comparators/RouteModelComparators.java
- **nit** [style] line 55 — `name()`/`creator()` swallow all exceptions as null; a narrower catch would still yield nulls-last ordering without masking unexpected failures, and `distance()`/`duration()` call `route.getUrl()` without the same guard, so the protection is inconsistent.

### route-converter-gui/src/main/resources/slash/navigation/converter/gui/RouteConverter_en.properties
- **minor** [docs] line 161 — The en (fallback) and de bundles add only the submenu key, not `route-length`/`route-duration`, unlike the other bundles; confirm those two column-header keys already exist in the base bundles, otherwise the Browse Length/Duration headers would render the raw keys.

---

cpesch:

**Reason:** Feedback is an inspector comment-only review (confidence 0.86) listing only minor/nit advisory findings, not a concrete change request. The one item with an implied code fix (FrameMenu mnemonic clash) is the PR author's own deliberately-scoped tradeoff (explicitly 'no invented translations' in the PR body) and fixing it would require inventing new labels/mnemonics across 23 locale bundles, outside patcher's authority. The docs item only asks to confirm route-length/route-duration keys already exist in the base bundles (they do, per RouteConverter_en.properties) — no edit needed. The comparator nit is pure style with no actionable instruction.
**Confidence:** 0.00

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-7694` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-7694).

PR #120, opened 2026-07-05
Wire BRouter into analyze mode for routed lengths (spec 00055 P2b)

cpesch:
Route-type (planned) position lists get realistic road lengths via BRouter-as-library instead of beeline; `--brouter-segments <dir>` now activates routing.

- Dependency: `org.btools:brouter-{core,expressions,mapaccess,codec,util}:1.7.8` — same coordinates/version the existing `brouter` module uses.
- `BRouterLegRouter` mirrors `slash.navigation.brouter.BRouter#getRouteBetween` (RoutingContext + RoutingEngine per leg), without the GUI DownloadManager machinery — routes against segments already on disk.
- Profile: `trekking` (bundled with its `lookups.dat` as module resources) — catalog routes are mostly cycling/hiking; documented in `analyze-json.md`.
- Fallback: all legs must route, else the whole list reports beeline (`lengthKind=beeline`); never a mix, never a crash — JSON always emitted. Track/Waypoint lists unchanged.
- Memory bounded (64 MB node cache, fresh engine per leg, per-leg timeout capped 60s) — fits the prod `-Xmx1g` analyzer.

Tests: 6 new in `BRouterRouteLengthComputerTest` + 1 end-to-end in `FileAnalyzerTest`; `mvn -pl route-converter-cmdline -am test` green.

Prod note: after merge the analyzer jar on vc612 gets rebuilt (`mvn -pl RouteConverterCmdLine -am package`) and re-shipped; Europe rd5 segments already at `/var/lib/brouter/segments` (87 tiles, 3 GB).

PR #122, opened 2026-07-05
Review follow-up: analyze mode hardening + BRouter single-pass routing (spec 00055)

cpesch:
Applies 8 review findings to the analyze/BRouter code merged in #120:

- **NaN/Infinity guard**: non-finite coordinates excluded from bbox union (all non-finite → `bbox:null`) — invalid-JSON output to the rc-site consumer is now impossible.
- **`lengthKind:null`** when no length is computable (was mislabeled `"track"` even for Route files); contract doc updated, Django side already tolerates null.
- **Routed<beeline sanity check**: physically impossible routed totals log a warning and fall back to beeline.
- **Single-pass routing**: one multi-waypoint `RoutingEngine` call per position list instead of one engine per leg — the rd5 node cache is no longer discarded N-1 times; seam renamed `LegRouter` → `RouteRouter`. Timeout scales with total beeline (cap 300s, documented).
- **Reuse over hand-rolling**: `Bearing.calculateBearing().getDistance()` replaces the equirectangular math; `InputOutput.copyAndClose` replaces the byte-copy loop; Jackson (`LinkedHashMap` + `ObjectMapper`) replaces the StringBuilder JSON assembly + `escape()` (~90 lines gone, output byte-identical — the exact-string contract test passes unchanged).
- **CLI tests**: new `RouteConverterCmdLineTest` covering all `--brouter-segments` branches incl. the dangling-option warning from the #120 review.

Tests: cmdline module 18 (7 FileAnalyzer, 7 BRouter computer, 4 CLI), full `-am` chain green.

### DIFF
--- a/route-converter-cmdline/src/main/doc/analyze-json.md
+++ b/route-converter-cmdline/src/main/doc/analyze-json.md
@@ -59,9 +59,12 @@ them from `bbox` centre via point-in-polygon (keeps geo data out of Java).
     routes successfully. The reported length is the sum of the on-road leg
     distances between consecutive route points.
   - **beeline** otherwise — no `--brouter-segments` given, the directory does
-    not exist, the list is outside coverage, or any single leg fails to route
-    (routing error, timeout, no segment). A partially-covered route is reported
-    wholly as `beeline`, never as a mix, so the label never over-promises.
+    not exist, the list is outside coverage, the route fails to route (routing
+    error, timeout, no segment), or the routed length comes back materially
+    shorter than the straight-line beeline through the same points (impossible
+    for an on-road route, so the result is distrusted). A partially-covered
+    route is reported wholly as `beeline`, never as a mix, so the label never
+    over-promises.
 
 Routing is best-effort and never aborts the run: any BRouter failure degrades
 that list to `beeline` and the JSON is still emitted.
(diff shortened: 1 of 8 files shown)


cpesch (commented):

Maintainer-authored review follow-up: single-pass BRouter routing, non-finite bbox guard, null `lengthKind`, and Jackson-based JSON emission. The change is clean, well-documented, and the new tests cover the routed/fallback, beeline-sanity, non-finite-bbox, null-length, and CLI-argument branches. No blockers; all convention checks pass. Two minor observations below.

### route-converter-cmdline/src/main/java/slash/navigation/converter/cmdline/BRouterRouteRouter.java
- **minor** [correctness] line 165 — Replacing the try-with-resources with `copyAndClose` leaks the open input stream if `Files.newOutputStream` throws before `copyAndClose` is entered (rare one-time-per-JVM extraction path).

### route-converter-cmdline/src/main/java/slash/navigation/converter/cmdline/FileAnalyzer.java
- **minor** [correctness] line 137 — Non-finite coordinates are now excluded from the bbox union, but the same NaN/Infinity values still flow into the length computer, where `round(NaN)` yields a misleading `lengthM` of 0 rather than a null; consider guarding the length path consistently.

---

PR #123, opened 2026-07-05
Review follow-up: crash telemetry privacy, robustness + consent UX (spec 00011)

cpesch:
Applies 10 review findings to the crash-telemetry feature merged in #118:

**Privacy**
- `DiagnosticReport.scrub` now redacts absolute-path-like tokens outside `user.home` (Windows drive paths, `/Volumes|/media|/mnt|/home/...`) from free-form messages — a `FileNotFoundException: D:\Trips\vacation2024.gpx` no longer leaves the machine. URLs and stack frames untouched; short usernames scrubbed via word-boundary matching.

**Robustness**
- Spool filenames gain the pid — two concurrent RC instances crashing in the same millisecond no longer overwrite each other's report.
- Free-text report fields capped at 24 KB at spool time — oversized reports could previously never be sent NOR deleted, re-POSTing every launch forever.
- One cycle-safe `ExceptionHelper.getCauseChain` now backs `getRootCause`, `getMessageWithCauses` and `DiagnosticReport` — an A→B→A cause cycle previously hung the crash handler itself.
- One-time WARNING when the committed placeholder HMAC secret is in use (misconfigured build was 100% silent report loss).

**Consent UX**
- Options dialog gains a "Send anonymous error reports" checkbox — consent is now revocable/grantable after the first-launch question (was one-shot forever). *Note: `.form` + generated UI edited in sync; worth one look in the IntelliJ designer.*
- First-launch consent dialog reformatted: bold headline + wrapped body (`<html><body width='420'>`) instead of one screen-wide paragraph; en + de.

**Test coverage**
- `MultipartRequest` raw-body path (the HMAC-verified bytes): entity type, exact UTF-8 bytes, precedence over multipart parts.
- `Application`'s uncaught-handler→`handleCrash` dispatch: called once, re-entry guard holds, latch resets.
- `CrashReporter` dialog-opener seam injected — manual-consent branching (claim latch, delete-only-when-sent) now unit-tested headless.

12 tests added; full reactor green.

### DIFF
--- a/common-gui/src/main/java/slash/navigation/gui/Application.java
+++ b/common-gui/src/main/java/slash/navigation/gui/Application.java
@@ -111,28 +111,30 @@ public abstract class Application {
         return bundle;
     }
 
-    public static <T extends Application> void launch(final Class<T> applicationClass, final List<String> bundleNames, final String[] args) {
-        // Last-resort net: since Java 7 the EDT routes uncaught exceptions to the
-        // thread's handler, which falls back to this default. Without it, a throw
-        // in an invokeLater/background runnable (e.g. the async map/profile view
-        // setup) died on System.err and never reached the log file or an error
-        // report — a blank/half-built UI with nothing to diagnose. Now every
-        // uncaught exception, on any thread, lands in the RC log.
-        //
-        // This also covers exceptions thrown on the EDT while a modal dialog's
-        // secondary event pump is active: on Java 9+ (verified on 17/21)
-        // EventDispatchThread.processException routes to
-        // getUncaughtExceptionHandler().uncaughtException(...), and the nested modal
-        // pump uses the same pumpOneEventForFilters -> processException path, so it
-        // falls back to this default handler too. No custom EventQueue is needed.
-        // (See EventDispatchThreadExceptionTest for the plain-EDT verification.)
+    /**
+     * Installs the last-resort uncaught-exception handler. Since Java 7 the EDT routes
+     * uncaught exceptions to the thread's handler, which falls back to this default.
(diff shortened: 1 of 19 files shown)


cpesch (commented):

Well-scoped review follow-up to the crash-telemetry feature (privacy scrubbing, spool-filename uniqueness, cycle-safe cause-chain traversal, revocable consent UX, HMAC placeholder warning) with 12 new tests; the .form and generated OptionsDialog.java are edited consistently. Two minor edge-case observations on DiagnosticReport scrubbing/truncation; no blockers.

### common/src/main/java/slash/common/log/DiagnosticReport.java
- **minor** [correctness] line 75 — The UNIX_PATH pattern matches a `/Volumes|media|mnt|home/…` segment anywhere in a string, so a URL such as `https://host/media/tiles/x.png` inside a message or stack trace is over-redacted to `<PATH>`; unlike the Windows pattern's backslash guard there is no unix equivalent, so the "URLs untouched" claim holds only for Windows paths.
- **nit** [correctness] line 59 — `MAXIMUM_FIELD_LENGTH` caps each field by character count, but the sender rejects by byte length; two fields of large multi-byte (e.g. CJK) text can still exceed the 64 KB byte limit and re-enter the never-sent/never-deleted loop this change targets.

---

cpesch:

**Reason:** Inspector's review verdict is explicitly comment-only ("no blockers") with only minor/nit-severity observations and no explicit change request from the maintainer — nothing actionable to patch.
**Confidence:** 0.0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-7723` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-7723).

PR #127, opened 2026-07-05
spec 00011: inject crash-report HMAC secret at build time (go-live prereq)

cpesch:
Prerequisite for activating Phase 2/3 crash telemetry. Without this, a shipped build signs every report with the `CHANGEME` placeholder and the server 403s them all.

Mirrors the existing `APIKeyRegistry` / `apikey.properties` pattern (Maven-filtered classpath resource) rather than per-launcher `-D` flags — so it covers the Mac app, Windows app, and `java -jar` identically, with no assembly/NSI changes.

## What changed
- **New filtered resource** `feedback/.../crash-report.properties` = `crashReportSecret=${crashReportSecret}`; `feedback/pom.xml` enables resource filtering.
- **`CrashReportSender.secret()` resolution order:** `routeconverter.crashReportSecret` system property (runtime override) → bundled resource (build-injected) → `CHANGEME` placeholder. An unresolved `${crashReportSecret}` token (local build) counts as unset via the same token-name guard `APIKeyRegistry` uses.
- **CI:** `_build-linux-mac.yml` and `_build-windows.yml` pass `-DcrashReportSecret=${{ secrets.CRASH_REPORT_HMAC_SECRET }}` at `mvn package`; reaches the reusable workflows via existing `secrets: inherit`.

## Verified
- `mvn -pl feedback test` green — 7 `CrashReportSenderTest` (2 new: unresolved-token→unset, system-property override).
- `-DcrashReportSecret=X process-resources` bakes `crashReportSecret=X`; a plain build leaves the token unresolved (→ placeholder). Confirmed both ways.

## Deploy coupling
Set the GitHub repo secret `CRASH_REPORT_HMAC_SECRET` on `cpesch/RouteConverter` to the **same** value as rc-site's `CRASH_REPORT_HMAC_SECRET`. The baked key is extractable from the shipped app by design (spec 00011).

### DIFF
--- a/.github/workflows/_build-linux-mac.yml
+++ b/.github/workflows/_build-linux-mac.yml
@@ -34,6 +34,8 @@ on:
     secrets:
       SIGNPATH_API_TOKEN:
         required: false
+      CRASH_REPORT_HMAC_SECRET:
+        required: false
 
 jobs:
   build:
--- a/.github/workflows/_build-linux-mac.yml
+++ b/.github/workflows/_build-linux-mac.yml
@@ -82,7 +82,7 @@ jobs:
         # value the inline jobs used before Phase 5.2.
         # No `clean` — runner is ephemeral; `clean` would also wipe the
         # Runtime-<arch> dirs unpacked from the hosted Mac JREs above.
-        run: ./mvnw --batch-mode -Prouteconverter -Dmaven.build.number=${{ github.run_number }} package
+        run: ./mvnw --batch-mode -Prouteconverter -Dmaven.build.number=${{ github.run_number }} -DcrashReportSecret=${{ secrets.CRASH_REPORT_HMAC_SECRET }} package
       - name: Verify runtime on stripped JRE (class-load gate)
         # Reproduces the minimized-JRE condition from scripts/jre-modules.txt and
         # fails the build on a ClassNotFoundException/NoClassDefFoundError that a
--- a/.github/workflows/_build-windows.yml
+++ b/.github/workflows/_build-windows.yml
@@ -70,7 +70,7 @@ jobs:
 
       - name: Build (incl. bundle modules, resource-filters .nsi)
         shell: cmd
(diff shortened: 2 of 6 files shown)


cpesch (commented):

Straightforward build-time secret-injection change that mirrors the existing `APIKeyRegistry` pattern (system-property override → bundled filtered resource → placeholder), and correctly treats both an unresolved Maven token and an empty CI secret as "unset" before falling back to the placeholder. No blockers. Two minor items worth a look before merge.

### feedback/src/test/java/slash/navigation/feedback/domain/CrashReportSenderTest.java
- **minor** [correctness] line 113 — `testSystemPropertyOverridesSecret` sets `SECRET_PROPERTY` via `System.setProperty` but never clears it, unlike the test above it. Surefire reuses one forked JVM across a module's test classes by default, so this leaks the override into any later test that calls `CrashReportSender.secret()` expecting default/bundled behavior — an order-dependent flake risk.
```suggestion
    @Test
    public void testSystemPropertyOverridesSecret() {
        System.setProperty(SECRET_PROPERTY, "a-real-injected-secret");
        try {
            assertEquals("a-real-injected-secret", CrashReportSender.secret());
        } finally {
            System.clearProperty(SECRET_PROPERTY);
        }
    }
```

### feedback/pom.xml
- **nit** [convention] line 44 — Filtering is enabled for all of `src/main/resources`, not scoped to `crash-report.properties`. Any other resource in this module with a literal `${...}` token (or a binary resource) would be silently rewritten/corrupted by the same filtering pass. Consider adding an `<includes>` scoped to the new file.

---

cpesch:
🤖 Patcher pushed a fix to `spec-00011-crash-secret-injection`.

cpesch (commented):
## Verdict: approve · confidence 0.92

Build-time HMAC secret injection mirrors the existing APIKeyRegistry filtered-resource pattern; resolution order (system property → bundled resource → placeholder) and the unresolved-token guard are correct and covered by new tests. No blockers found in this second pass; only cosmetic nits remain, so this closes out the polish round.

---

PR #128, opened 2026-07-05
fix(feedback): crash-secret tests fail under CI build-time injection (master hotfix)

cpesch:
**Master prereleases are red.** #127 merged the build-time HMAC injection, but two `CrashReportSenderTest` cases assumed the `${crashReportSecret}` token stays unresolved. CI's `mvn package -DcrashReportSecret=<key>` filters the main `crash-report.properties` with the real key, so `bundledSecret()`/`secret()` return it and `testBundledSecretUnresolvedTokenTreatedAsUnset` + `testWarnsOnceWhenSecretIsPlaceholder` fail — on `build-linux-mac` and `build-windows` (reactor stops at `:feedback`). Local `mvn test` (no `-D`) passed, which is why it slipped through.

## Fix
Add a **test-scoped** `feedback/src/test/resources/.../crash-report.properties` keeping the literal token. Test resources aren't filtered and `target/test-classes` precedes `target/classes`, so it shadows the injected main resource — the unit tests always see the not-injected state, independent of whether the packaging build baked a key.

## Verified
`mvn -pl feedback -DcrashReportSecret=SIMULATED-CI-VALUE test` → green (reproduces the CI condition). Confirmed main resource = `SIMULATED-CI-VALUE`, test resource = `${crashReportSecret}`.

### DIFF
--- a/feedback/src/test/resources/slash/navigation/feedback/domain/crash-report.properties
+++ b/feedback/src/test/resources/slash/navigation/feedback/domain/crash-report.properties
@@ -1,0 +1,6 @@
+# Test-scoped copy with the token deliberately left UNRESOLVED. Test resources
+# are not filtered, and target/test-classes precedes target/classes on the test
+# classpath, so this shadows the main resource. That keeps the unit tests seeing
+# the "secret not injected" state even when the CI `mvn package` build filters
+# the main resource with -DcrashReportSecret=<real key>.
+crashReportSecret=${crashReportSecret}


cpesch (commented):

Test-scoped resource file correctly exploits Maven's default test-classpath ordering (`target/test-classes` before `target/classes`) to keep the two `CrashReportSenderTest` cases hermetic under `-DcrashReportSecret` packaging builds; the PR description's verification run substantiates the claim. One minor maintainability note below.

### feedback/src/test/resources/slash/navigation/feedback/domain/crash-report.properties
- **nit** [convention] line 6 — Test resource duplicates the main resource's key name with no automated check that the two stay in sync if the main properties key is ever renamed.

---

cpesch:

**Reason:** Feedback verdict is explicitly 'comment-only' and the sole item is labeled 'nit' — an optional observation, not a change request. No concrete instruction to act on.
**Confidence:** 0.3

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-7811` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-7811).

PR #130, opened 2026-07-05
Local catalog: hide dotfiles, resolve macOS Finder aliases, swap LNK parser to mslinks

cpesch:
Three related improvements to browsing the **local route catalog**:

### 1. Hide dotfiles (e.g. `.DS_Store`)
New `RouteFileFilter` excludes dot-prefixed files from the local route list, mirroring the existing dot-directory rule in `DirectoryFileFilter`. Previously `.DS_Store` showed up as a route on macOS.

### 2. macOS Finder alias support
New `slash.common.io.MacAlias` — the counterpart to `WindowsShortcut`:
- detection via the `book` bookmark-data magic (cheap, pure Java), guarded by `Platform.isMac()`
- resolution via `osascript` (Finder → *original item* → POSIX path), following the `BabelFormat` external-process precedent

Wired into `LocalCategory.getCategories()`/`getRoutes()` and `DirectoryFileFilter`, parallel to the `WindowsShortcut` branches. Handles **both** file-aliases (→ routes) and directory-aliases (→ categories).

### 3. `WindowsShortcut` → mslinks
Replaced the 2011 hand-rolled LNK byte-parser with `com.github.vatbub:mslinks:1.0.6.2` (permissive license, no runtime transitive deps). The `isPotentialValidLink` / `isDirectory` / `isFile` / `getRealFilename` facade is unchanged, so `LocalCategory` and `MacAlias` are unaffected. More robust on Unicode / UNC / relative targets.

### Tests
- `WindowsShortcutTest` — reads a real `[MS-SHLLINK]` spec fixture, asserts target `C:\test\a.txt` + file/dir flags
- `MacAliasTest` — creates real Finder aliases via `osascript` (auto-skips on non-mac / headless via JUnit `Assume`)
- existing `route-catalog` local-catalog tests: 42/42, no regression


cpesch:

Solid, well-tested change (dotfile hiding, mslinks swap, macOS alias support). One robustness issue is repeated in two spots and is worth fixing before merge; the rest are minor/nit. All convention checks pass.

### route-catalog/src/main/java/slash/navigation/routes/local/LocalCategory.java
- **major** [correctness] line 143 — In `getRoutes()`, a dangling/unresolvable alias makes `new MacAlias(file)` throw `IOException`, which propagates and aborts the entire route listing. Wrap it in try/catch and `continue` on failure, like `DirectoryFileFilter.isAlias()` already does.
- **major** [correctness] line 89 — Same failure mode in `getCategories()` for directory aliases: one dangling alias aborts the whole category listing. Guard the construction and skip on `IOException`.

### common/src/main/java/slash/common/io/MacAlias.java
- **minor** [correctness] line 110 — `resolveTarget()` does not destroy the `osascript` process if `readLine()` throws `IOException` (only `InterruptedException` is handled), leaking the child; the error stream is also never drained, which can stall on a large stderr.
- **nit** [docs] line 48 — `isPotentialValidAlias()` checks only the 4-byte `book` prefix though the class doc cites the fuller magic; the `@return` line is also ungrammatical.

### common/src/main/java/slash/common/io/WindowsShortcut.java
- **minor** [correctness] line 76 — The public `isLocal()` method was dropped in the mslinks swap; confirm no remaining callers outside this diff.

---

cpesch:
🤖 Patcher pushed a fix to `feature/local-catalog-aliases`.

PR #131, opened 2026-07-05
ci(release): prune stale assets from /latest/ and /prerelease/ with rsync --delete

cpesch:
Renamed/discontinued assets (e.g. 3.4 `*Bundle.exe`, `*OpenSource.*`) lingered in the release/prerelease download dirs because both deploy rsyncs ran without `--delete`. After 3.5 renamed the artifacts to single `RouteConverterWindows.exe` / `RouteConverterLinux.jar` / `RouteConverterMac-{aarch64,x64}-app.zip`, the old files stayed in `/latest/` and the download page kept serving the stale 3.4 Windows build under its old name (forum thread-4133).

Add `--delete` to both deploy rsyncs so each publish leaves only the current release/prerelease set. Matches the javadoc rsync, which already uses `--delete`.

## Changes
- `.github/workflows/release.yml` — `--delete` on the `/latest/` rsync
- `.github/workflows/prerelease.yml` — `--delete` on the `/prerelease/` rsync

## Note
Live cleanup (WP downloads page, releases-host landing, PAD catalog, orphan file removal) already done out-of-band; this PR is the durable fix that prevents recurrence.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -71,7 +71,9 @@ jobs:
         env:
           DEPLOY_HOST: ${{ secrets.RC_RELEASE_DEPLOY_HOST }}
         run: |
-          rsync -av --omit-dir-times --no-perms \
+          # --delete: /prerelease/ mirrors only the current build; prunes renamed/
+          # discontinued assets so stale files aren't served (matches release.yml)
+          rsync -av --delete --omit-dir-times --no-perms \
             artefacts/ \
             "rc-release-deploy@${DEPLOY_HOST}:/var/www/routeconverter.com/static/downloads/prereleases/"
       - name: Update rolling 'prerelease' GitHub Release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,7 +79,9 @@ jobs:
         env:
           DEPLOY_HOST: ${{ secrets.RC_RELEASE_DEPLOY_HOST }}
         run: |
-          rsync -av --omit-dir-times --no-perms \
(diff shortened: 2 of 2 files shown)


cpesch:

Adds `--delete` to the `/latest/` and `/prerelease/` deploy rsyncs so each publish prunes stale/renamed assets (fixes forum thread-4133), matching the existing javadoc rsync. Fix direction is right, but `--delete` makes the step destructive -- flagging two points to confirm before merge.

### .github/workflows/release.yml
- **major** [correctness] line 84 — `--delete` now means a partial artefacts/ dir (matrix build failure, or an artifact not yet downloaded when this step runs) will prune a still-valid previous-version file for that platform with nothing to replace it — regressing the download page the same way this PR fixes, just from the opposite direction. Confirm the job hard-depends on successful build+download of every platform artifact (Windows/Linux/Mac/Portable) before this step runs, or add a sanity check on the expected file count in `artefacts/` first.
- **minor** [correctness] line 82 — "Matches the javadoc rsync" is good precedent, but only holds if javadoc publishing goes through the same `rc-release-deploy` restricted-rsync wrapper/user. Worth a one-line confirmation that this wrapper's allowed-options list actually accepts `--delete` for this destination (some forced-command rrsync setups whitelist a narrower option set).

### .github/workflows/prerelease.yml
- **major** [correctness] line 76 — same partial-build risk as release.yml: with `--delete`, an incomplete `artefacts/` dir at rsync time prunes a prior valid file instead of leaving it in place.
- **minor** [correctness] line 74 — same wrapper-precedent question as release.yml.

---

PR #132, opened 2026-07-05
Fix format detection of remote catalog files larger than 1 MB

cpesch:
### Symptom
Opening a catalog route by its HTTP URL fails with **"Konnte '…' nicht öffnen, da das Dateiformat nicht erkannt wurde."** for files larger than ~1 MB — e.g. a 1.5 MB Columbus GPS track (`api.routeconverter.com/files/<uuid>`). The same file opens fine as a local file, and `RouteConverterCmdLine analyze` detects it correctly.

### Root cause
`read(URL)` sizes the `mark()`/`reset()` buffer from `getSize(URL)`, which for non-`file` URLs returned a fixed `TOTAL_BUFFER_SIZE` (1 MB) instead of the real content length. During format probing, a format that scans the whole stream (tried *before* the matching one in registry order — e.g. NMEA before Columbus) reads past the mark limit, so `reset()` throws *"Resetting to invalid mark"* and every later format — including the correct one — never sees the data. Local files work because their size comes from `File.length()`.

This affects **any** catalog file over 1 MB, not just Columbus (spec 00055 measured live sizes p99 ≈ 6 MB, max 15 MB).

### Fix
Both URL read paths now read the response fully into memory once and parse from a `ByteArrayInputStream` sized to the actual content, so `reset()` between format attempts always succeeds. Also removes the redundant second HTTP request (`getSize` + `openStream`) and the now-unused `getSize(URL)`.

### Test
`NavigationFormatParserUrlSizeTest` serves a >1 MB body over a local `HttpServer` (with `Content-Length`) and asserts the format is detected — fails before the fix, passes after. Existing Columbus + base parser tests remain green.


cpesch (commented):

Root-cause fix is sound: reading the full URL response into memory and sizing mark()/reset() from the actual byte count correctly resolves the reset() failure on large catalog files, and the new HttpServer-backed regression test directly exercises the failure mode. Only two trivial style nits found; no correctness, security, or convention issues.

### navigation-formats/src/main/java/slash/navigation/base/NavigationFormatParser.java

- **nit** [style] line 223 — Wrapping the in-memory ByteArrayInputStream in a BufferedInputStream(..., CHUNK_BUFFER_SIZE) is no longer needed since the full content is already buffered in the byte array.
- **nit** [style] line 225 — The '+ CHUNK_BUFFER_SIZE * 2' padding on the mark() read limit is a leftover from the old size-guessing logic; bytes.length alone already covers the full available content and would be clearer.

---

PR #134, opened 2026-07-06
docs(spec 00009): retire — no low-maintenance bundle-size win

cpesch:
Closes spec 00009 (reduce bundle download size generically) as **retired** — investigated all four options, none yields a low-maintenance, version-agnostic shrink.

## Findings (full detail in the spec)
- **Option 2 `minimizeJar`**: 61.5M → 16.4M on the Linux fat jar, but strips reflection-loaded classes (Garmin FIT 550→0, JavaHelp, jfreechart, JAXB, own format parsers) → app broken. Rejected.
- **Option 1a jlink JRE trim**: already shipped (`jre-modules.txt` + `verify-runtime.sh`). Nothing to add.
- **Option 1b jdeps app-jar closure**: better-rooted than minimizeJar (formats survive), but inherits the same reflection keep-rule burden. Not worth reimplementing a shade-minimizer.
- **Option 4 exclusions**: trialled osmosis/protobuf, Rhino, hppc, jts — all live at runtime (`PbfUtil` bbox, proxy-vole PAC, graphhopper routing/geometry). No dead whole-jars.

**Conclusion**: JRE already trimmed; app-jar weight is genuine. Further shrink needs per-class keep-rules (rejected maintenance cost) or a feature-level decision (e.g. drop PBF import → sheds protobuf+osmosis). Reopen only then.

## Also
- Fixes invalid `status: closed` on spec 00008 → `shipped` (was breaking `scripts/gen-status.py`).

Docs-only. No code or build changes.


cpesch (commented):

Docs-only retirement of spec 00009 (git mv into `specs/done/` with empirical findings) plus a valid 00008 status fix (`closed` → `shipped`) by the maintainer. Content is coherent and there are no blockers; one minor status/dir consistency note surfaced by the STATUS.md regeneration.

### specs/STATUS.md
- **minor** [docs] line 31 — The regenerated Done table lists 00015 as `shipped` but with a `./` (active-dir) path instead of `./done/`, unlike every other done row. The file appears not to have been `git mv`'d into `specs/done/` — a status/dir mismatch `gen-status.py` warns on. Out of this PR's scope, but worth a follow-up move.

---

PR #145, opened 2026-07-09
Fix color-picker jank (spec 00015) + add reset-to-default color buttons

cpesch:
Two independent changes, one commit each.

## 1. `fix(mapview)`: coalesce non-selected-list rebuilds — unjank color picking
Spec 00015 wired `NonSelectedPositionListsRenderer.update()` into the map view's `repaintPositionListListener`, which fires on every route/track/waypoint **color** and **line-width** change. `update()` rebuilt the whole gray overlay **synchronously on the EDT** (`invokeInAwtEventQueue` runs inline when already on the EDT), so a `JColorChooser` drag — a flood of change events — did one full clear+repopulate *per event*. With a multi-list file loaded, the Options color dialog janked.

Fix: schedule a **single deferred rebuild** on the EDT (`AtomicBoolean` guard + `invokeLater`), dropping redundant pending requests. The rebuild reads the latest model state, so collapsing several into one loses nothing — mirroring the decoupling the selected-route rebuild already gets from `UpdateDecoupler`.

**Tests:** two regression tests — a 50-event drag goes from **50 rebuilds → 1**; both fail on the pre-fix inline version, pass with the fix.

## 2. `feat(options)`: reset-to-default color button per chooser
Each of the Route/Track/Waypoint color choosers gets a right-aligned **"Reset to default color"** button directly below it. Clicking it sets the chooser to the model's built-in default, which propagates to the `ColorModel` — and on to the map — via the existing selection listener.
- `ColorModel.getDefaultColor()`
- `OptionsDialog.setupColorChooser(chooser, model)` — wraps each chooser in place, button beneath
- new `reset-color` bundle key (EN + DE)

## Validation
- `mapsforge-mapview` suite green on JDK 21 (7/7 incl. the 2 new tests); negative check confirms the tests catch the regression.
- Full reactor packages successfully.
- Launched the app (11-track sample): color-drag is smooth; reset buttons snap each chooser to its default and update the map.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/NonSelectedPositionListsRenderer.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/NonSelectedPositionListsRenderer.java
@@ -36,8 +36,9 @@ import slash.navigation.mapview.mapsforge.lines.Polyline;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
-import static slash.common.helpers.ThreadHelper.invokeInAwtEventQueue;
+import static javax.swing.SwingUtilities.invokeLater;
 import static slash.navigation.base.RouteCharacteristics.Route;
 import static slash.navigation.base.RouteCharacteristics.Waypoints;
 import static slash.navigation.mapview.mapsforge.helpers.ColorHelper.asRGBA;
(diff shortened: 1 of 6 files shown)


cpesch (commented):

Two independent, well-scoped changes: an EDT-coalescing fix for the non-selected-list renderer (regression-tested, matches the described spec 00015 jank) and a reset-to-default-color button added via a small setupColorChooser refactor. Logic, threading, and i18n additions all check out; no convention-check violations found.

The `AtomicBoolean` guard + `invokeLater` coalescing in `NonSelectedPositionListsRenderer.update()` correctly collapses rapid change events into one deferred rebuild, and the two new tests (`updateDefersTheRebuildInsteadOfRunningItOnTheEventThread`, `rapidUpdatesCoalesceIntoASingleRebuild`) exercise exactly that regression via `invokeAndWait` draining. The `OptionsDialog.setupColorChooser` refactor cleanly dedupes the three near-identical chooser setups and wraps each in place for the new reset button, propagating through the existing `ColorModel` selection listener. No blockers found.

---

PR #146, opened 2026-07-09
Fix Java-update dialog linking to outdated Adoptium (Java 17) URL

cpesch:
The in-app "a newer Java is available" dialog hardcoded
`https://adoptium.net/de/temurin/releases?version=17` — so it pointed users at
**Java 17** even while offering a newer **Java 21** build, and forced the German
(`/de/`) locale regardless of the user's language.

Fix: derive the major version from the Java version actually being offered
(`result.getLatestJavaVersion()`), so the link always matches the dialog, and drop
the hardcoded locale. This self-updates when the toolchain retargets to Java 25.

```java
int javaMajor = new Version(result.getLatestJavaVersion()).getMajor();
showUpdateMessage(window, message, "https://adoptium.net/temurin/releases/?version=" + javaMajor);
```

Reported on the forum (companion to the website download-page link fix, which is already live).

### DIFF
--- a/route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/UpdateChecker.java
+++ b/route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/UpdateChecker.java
@@ -199,7 +199,8 @@ public class UpdateChecker {
     private void offerJavaUpdate(Window window, UpdateResult result) {
         String message = format(BaseRouteConverter.getBundle().getString("confirm-java-update"),
                 result.getMyJavaVersion(), result.getLatestJavaVersion());
-        showUpdateMessage(window, message, "https://adoptium.net/de/temurin/releases?version=17");
+        int javaMajor = new Version(result.getLatestJavaVersion()).getMajor();
+        showUpdateMessage(window, message, "https://adoptium.net/temurin/releases/?version=" + javaMajor);
     }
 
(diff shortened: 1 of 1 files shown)


cpesch (commented):

Correctly derives the Adoptium major version from the Java version actually being offered and drops the hardcoded `/de/` locale, matching the described bug and staying self-updating for future JDK bumps. One thing worth confirming before merge: the diff doesn't show an import change, so double-check `Version` is already available in this file.

### route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/UpdateChecker.java

- **minor** [correctness] line 202 — The new `new Version(result.getLatestJavaVersion()).getMajor()` call needs `Version` in scope, but the diff only touches the method body and doesn't add an import — confirm `Version` is already imported/used elsewhere in this file (e.g. for other version comparisons) so this compiles, and that CI is green.

---

cpesch:

**Reason:** No code change needed: slash.common.system.Version is already imported (see existing `import slash.common.system.Version;`) and used elsewhere in the file (e.g. isLaterVersionThan/getLatestJavaVersion), so the inspector's compile-safety concern is already satisfied.
**Confidence:** 0.95

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-8885` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-8885).

PR #153, opened 2026-07-13
fix(i18n): English fallback for keys missing in the active locale (fixes Save As crash)

cpesch:
## Symptom
Save As (and any dialog going through `DialogStrings.asDialogString`) crashes on a non-English UI:

```
MissingResourceException: Can't find resource for bundle
  slash.navigation.gui.helpers.CombinedResourceBundle, key list-none
    at DialogStrings.asDialogString(DialogStrings.java:54)
    at FileOperations.saveFiles(FileOperations.java:216)
    at ...SaveAsAction.run(SaveAsAction.java:42)
```

## Root cause
`CombinedResourceBundle.load()` copied keys only from the **active-locale** bundle. English lives in `*_en.properties` and there is **no no-suffix base bundle**, so a key present only in the English bundle has no fallback → `MissingResourceException` for every non-English locale. The `list-none` / `list-and` / `list-more` keys (added with spec 00016, consumed by `DialogStrings`) existed only in `RouteConverter_en.properties`, so every German/French/… user crashed on Save As.

## Fix
- `CombinedResourceBundle.load()`: load English first as a fallback layer, then overlay the active locale. Any key missing in the active locale now degrades to English instead of throwing. This closes the whole class of "translated-in-en-only" crashes, not just `list-*`.
- Add the German translations for the three `list-*` keys (`keine` / `und` / `und {0} weitere`) so the German UI shows German, not the English fallback.
- Regression test `englishActsAsFallbackForKeysMissingInTheActiveLocale` (+ `testbundle3_en`/`_de` fixtures). It reproduces the exact `MissingResourceException` on the pre-fix loader and passes with the fix.

## Validation
`common-gui` `CombinedResourceBundleTest` — 6/6 green on JDK 21; verified the new test fails (same exception as the report) when the fix is reverted.

### DIFF
--- a/common-gui/src/main/java/slash/navigation/gui/helpers/CombinedResourceBundle.java
+++ b/common-gui/src/main/java/slash/navigation/gui/helpers/CombinedResourceBundle.java
@@ -39,11 +39,18 @@ public class CombinedResourceBundle extends ResourceBundle {
 
     public void load() {
         bundleNames.forEach(bundleName -> {
-            ResourceBundle bundle = getBundle(bundleName);
-            Enumeration<String> keysEnumeration = bundle.getKeys();
-            ArrayList<String> keysList = list(keysEnumeration);
-            keysList.forEach(key -> resources.put(key, bundle.getString(key)));
+            // Load English first as a fallback, then let the active-locale bundle override
+            // it. There is no no-suffix base bundle (English lives in *_en.properties), so a
+            // key present only in the English bundle would otherwise have no fallback and
+            // throw MissingResourceException for every other locale - e.g. the list-none /
+            // list-and / list-more keys used by DialogStrings on a German UI.
+            merge(getBundle(bundleName, Locale.ENGLISH));
+            merge(getBundle(bundleName));
         });
+    }
+
+    private void merge(ResourceBundle bundle) {
+        list(bundle.getKeys()).forEach(key -> resources.put(key, bundle.getString(key)));
     }
 
     public Object handleGetObject(String key) {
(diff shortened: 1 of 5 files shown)


cpesch:

The English-fallback fix correctly resolves the reported Save As crash and is well covered by a regression test, but the refactor changes merge ordering in a way that can let a later bundle's English value overwrite an earlier bundle's correct active-locale translation for a shared key — a variant of the exact bug class this PR fixes.

### common-gui/src/main/java/slash/navigation/gui/helpers/CombinedResourceBundle.java

- **major** [correctness] line 49 — Merging is interleaved per bundleName (English then active-locale for bundle N, then English then active-locale for bundle N+1). `CombinedResourceBundleTest.getKeysExposesEveryMergedKey` confirms this class is used with multiple bundleNames sharing keys; if a later bundleName's English variant defines a key that an earlier bundleName already translated into the active locale, and the later bundleName's own active-locale variant doesn't define that key, the later English value clobbers the earlier correct translation — reintroducing an English-instead-of-translated-text bug across bundle boundaries.
```suggestion
public void load() {
    bundleNames.forEach(bundleName -> merge(getBundle(bundleName, Locale.ENGLISH)));
    bundleNames.forEach(bundleName -> merge(getBundle(bundleName)));
}
```
- **minor** [correctness] line 48 — `getBundle(bundleName, Locale.ENGLISH)` throws `MissingResourceException` if that bundleName has no `*_en` properties file and no no-suffix base file. This is a shared common-gui class (also used by TimeAlbumPro per the base-module docs), so any consumer whose bundle list lacks an English variant would newly crash on `load()` where it previously loaded fine off the active locale alone.

---

PR #156, opened 2026-07-14
Validate downloads against any known-good checksum (#155)

### LINKED ISSUE #155: Validate downloads against any known-good checksum (BRouter segments perpetually fail catalog checksum)
## Summary

BRouter routing segments (and, in principle, every `Copy`/`Extract` datasource file) **perpetually fail checksum validation** on the client. RouteConverter points clients directly at the live upstream (`brouter.de`) for the download, but validates the downloaded bytes against a **single snapshot checksum** in the RouteConverter catalog. Because `brouter.de` rebuilds its segment tiles nightly, the downloaded bytes almost never match the catalog snapshot — so RouteConverter logs a checksum warning and re-queues the (identical, from its point of view) download on every launch, a loop that never settles.

The data model was designed to avoid exactly this (a file may carry **several** `<checksum>` entries), but the read side collapses that list to a single "latest" checksum and the write side only ever stores one. This issue fixes it: **validate a downloaded file against _any_ known-good checksum for that file**, and have the catalog generator **accumulate** the set.

## Evidence

Real user report (feedback #1323): offline BRouter routing in southern Spain, `W5_N35.rd5`. The log shows RouteConverter flagging the file as invalid on startup, then routing works anyway (the file is fine), and later a genuine `target island` on one leg — a *separate* data-gap problem. The checksum warning is noise:

```
Validator isChecksumValid: .../brouter/segments4/W5_N35.rd5 has 61231383 bytes but expected 61235590
Validator isChecksumValid: .../brouter/segments4/W5_N35.rd5 has SHA-1 E77CFCC…FAE0E but expected 3191EAC…E70A4
DownloadManager scanForOutdatedFilesInQueue: Found outdated download … W5_N35.rd5
```

Four different builds of the *same* tile were observed within ~10 days:

| Source | Build | contentLength |
|---|---|---|
| User's local file | ~Jul 5 | 61,231,383 |
| Catalog *expected* at that time | ~Jul 5 | 61,235,590 |
| Catalog advertised now | 2026-07-13 01:03Z | 61,273,719 |
| brouter.de serving now | 2026-07-14 01:03Z | 61,279,809 |

The catalog datasource points straight at upstream:

```
<datasource id="brouter-segments-4" name="BRouter Segments"
            baseUrl="http://brouter.de/brouter/segments4/" directory="brouter/segments4" action="Copy">
```

So the download source (live, nightly) and the checksum authority (catalog snapshot) are two systems on different cadences — mismatch is structural, not a corrupted file.

## Root cause — three layers

1. **Schema already supports a set** ✅ — `DownloadableType.checksum` is `List<ChecksumType>`, `maxOccurs="unbounded"`; `Downloadable.getChecksums()` exposes the list.
2. **Client collapses the set to one** ❌ — every consumer builds `FileAndChecksum` with `downloadable.getLatestChecksum()` (newest by `lastModified`): `BRouter.downloadSegment` (`BRouter.java:417`), `DataSourceManager.addOrUpdateInQueue`/`queueForDownload`/`asFragments` (`DataSourceManager.java:248,261,274`), `DataSourceImpl.getDownloadableBySHA1` (`DataSourceImpl.java:132,142`). `Validator.isChecksumValid` then compares `actual` against that single `getExpectedChecksum()`.
3. **Generator writes one, replacing** ❌ — `UpdateCatalog.updateFile/updateMap/updateTheme` calls `createFileType(uri, singletonList(checksum), …)`; the served catalog carries exactly one `<checksum>` per file.

## Fix (this PR)

**Validate against any known-good checksum, and accumulate the set.**

1. `FileAndChecksum` carries a **list** of expected checksums (single-arg constructor kept for back-compat; `getExpectedChecksum()` still returns the latest for display/size).
2. `Validator.isChecksumValid` accepts the file if `actual` matches **any** expected checksum (reusing the existing per-checksum SHA-1/size logic); warnings only when *no* candidate matches.
3. Consumers pass `downloadable.getChecksums()` (the list) instead of `getLatestChecksum()` — `BRouter.downloadSegment` first (the reported case), plus the generic `DataSourceManager` paths and fragments.
4. `UpdateCatalog` **appends** each newly-observed checksum to the file's existing `getChecksums()` (dedup, newest-first, capped to the last N) instead of `singletonList`, so the catalog accumulates recent upstream builds.

### Limitation
The checksum-set tolerates upstream drift only for builds the catalog has already scanned; a brand-new build downloaded before the next `UpdateCatalog` run still mismatches. Running `UpdateCatalog` on a cadence (e.g. daily, after upstream's rebuild) keeps the set current.

### Airtight alternative (not in this PR)
Mirror the segments onto RouteConverter infrastructure and generate the catalog **from the mirrored files** (`baseUrl` → the mirror). Checksum and served bytes are then the same build by construction, independent of date; `MirrorCatalog` already exists for this. Costs storage + a sync job. The two approaches compose.

_Filed from RouteConverter feedback error report #1323._

cpesch:
Closes #155.

BRouter segment tiles perpetually failed catalog checksum validation. The catalog points clients straight at `brouter.de` (which rebuilds tiles nightly), but validates the downloaded bytes against a single, older catalog snapshot — so every launch re-flagged the file as outdated and re-queued it (non-fatal, but noisy + wasteful). Surfaced by feedback error report #1323.

The schema already allowed several checksums per file (`DownloadableType.checksum` is unbounded), but the read side collapsed the list to `getLatestChecksum()` and the generator only wrote `singletonList()`.

### Changes
- **`FileAndChecksum`** carries a list of expected checksums via a `forChecksums()` static factory; the single-`Checksum` constructor is kept so the many `(file, null)` call sites stay unambiguous. `getExpectedChecksum()` still returns the latest (for size/display).
- **`Validator`** accepts a download if its actual checksum matches **any** known-good entry; the per-checksum SHA-1/size logic (incl. the spec 00054 "later than remote" heuristic) is extracted into `matches()`.
- **Consumers** pass the full list: `BRouter` (segments + profiles), `DataSourceManager` (files + fragments), `GraphHopper`; `DataSourceImpl` SHA-1 lookups scan all checksums.
- **`UpdateCatalog`** appends each newly-observed checksum to the file's existing set (dedup, newest-first, capped at 30) instead of replacing with one.

Single-element lists behave exactly as before, so mirror-served datasources are unaffected.

### Tests
`ValidatorTest` gains match-any / match-none / empty-list cases. Green across `download`, `datasource`, `brouter`, `download-tools`, `graphhopper` (unit tests; ITs not run).

### Not in this PR
The checksum-set only tolerates builds `UpdateCatalog` has already scanned; running it on a cadence keeps the set current. The airtight alternative — mirror the segments onto RouteConverter infra and generate the catalog from the mirrored files (via the existing `MirrorCatalog`) — is noted in #155.

### DIFF
--- a/brouter/src/main/java/slash/navigation/brouter/BRouter.java
+++ b/brouter/src/main/java/slash/navigation/brouter/BRouter.java
@@ -414,7 +414,7 @@ public class BRouter extends BaseRoutingService {
         String uri = downloadable.getUri();
         String url = getSegmentsBaseUrl() + uri;
         return downloadManager.queueForDownload(getName() + " Routing Segment: " + uri, url, Action.valueOf(getSegments().getAction()),
-                new FileAndChecksum(createSegmentFile(downloadable.getUri()), downloadable.getLatestChecksum()), null);
+                FileAndChecksum.forChecksums(createSegmentFile(downloadable.getUri()), downloadable.getChecksums()), null);
     }
 
     public void downloadRoutingData(List<MapDescriptor> mapDescriptors) {
@@ -523,7 +523,7 @@ public class BRouter extends BaseRoutingService {
         String uri = downloadable.getUri();
         String url = getProfilesBaseUrl() + uri;
         downloadManager.queueForDownload(getName() + " Routing Profile: " + uri, url, Action.valueOf(getProfiles().getAction()),
-                new FileAndChecksum(createProfileFile(downloadable.getUri()), downloadable.getLatestChecksum()), null);
+                FileAndChecksum.forChecksums(createProfileFile(downloadable.getUri()), downloadable.getChecksums()), null);
     }
 
     public void downloadProfiles() {
--- a/datasource/src/main/java/slash/navigation/datasources/DataSourceManager.java
+++ b/datasource/src/main/java/slash/navigation/datasources/DataSourceManager.java
@@ -245,7 +245,7 @@ public class DataSourceManager {
 
         downloadManager.addOrUpdateInQueue(dataSource.getName() + ": " + downloadable.getUri(),
                 dataSource.getBaseUrl() + downloadable.getUri(), action,
-                new FileAndChecksum(target, downloadable.getLatestChecksum()),
+                FileAndChecksum.forChecksums(target, downloadable.getChecksums()),
                 asFragments(directory, downloadable.getFragments(), action.equals(Extract)));
     }
 
@@ -258,7 +258,7 @@ public class DataSourceManager {
 
         return downloadManager.queueForDownload(dataSource.getName() + ": " + downloadable.getUri(),
                 dataSource.getBaseUrl() + downloadable.getUri(), action,
-                new FileAndChecksum(target, downloadable.getLatestChecksum()),
+                FileAndChecksum.forChecksums(target, downloadable.getChecksums()),
                 asFragments(target, downloadable.getFragments(), false));
     }
 
@@ -271,7 +271,7 @@ public class DataSourceManager {
         for (Fragment<Downloadable> fragment : fragments) {
             String key = fragment.getKey();
             File target = new File(useDirectoryParent ? directory.getParentFile() : directory, key);
-            result.add(new FileAndChecksum(target, fragment.getLatestChecksum()));
+            result.add(FileAndChecksum.forChecksums(target, fragment.getChecksums()));
         }
         return result;
     }
--- a/datasource/src/main/java/slash/navigation/datasources/impl/DataSourceImpl.java
+++ b/datasource/src/main/java/slash/navigation/datasources/impl/DataSourceImpl.java
@@ -128,21 +128,19 @@ public class DataSourceImpl implements DataSource {
 
     public Downloadable getDownloadableBySHA1(String sha1) {
         initialize();
-        for(Downloadable downloadable : downloadableMap.values()) {
-            Checksum checksum = downloadable.getLatestChecksum();
-            if(checksum != null && sha1.equals(checksum.getSHA1()))
-                return downloadable;
-        }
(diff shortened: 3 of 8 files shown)


cpesch:

Solid, well-tested fix for the core match-any-checksum problem — the single-checksum path is unchanged and the new match-any path has good targeted tests. Two issues worth addressing before merge: a null-short-circuit bug in the new multi-checksum loop that can silently disable checksum validation, and untested dedup logic in the catalog updater.

### download/src/main/java/slash/navigation/download/actions/Validator.java
- **major** [correctness] line 119 — `matches()` returns `true` unconditionally when `expected` is `null` (see line 133-134). Since `isChecksumValid()` now loops over the full `expectedChecksums` list looking for any match, a single `null` entry anywhere in that list short-circuits the loop to always-valid, bypassing the real checksums entirely. This PR's own `DataSourceImpl` changes add defensive `checksum != null` guards for the same underlying lists (`downloadable.getChecksums()`/`fragment.getChecksums()`), suggesting null entries are a real, anticipated case here too.
```suggestion
        for (Checksum expected : expectedChecksums)
            if (expected != null && matches(expected, actual)) {
                log.fine(format("%s has valid checksum", file.getFile()));
                return true;
            }
```

- **minor** [correctness] line 122 — `mergeChecksums()`'s de-duplication (`!result.contains(checksum)`) relies on `Checksum` overriding `equals()`/`hashCode()` by value; the freshly-observed checksum is always a new object, so reference-equality `contains()` would never dedupe in practice. No test in this PR exercises `mergeChecksums()` directly to confirm the dedup claim in the PR description holds — worth adding a small `UpdateCatalogTest` case.

---

cpesch:
🤖 Patcher pushed a fix to `fix/155-validate-any-checksum`.

PR #162, opened 2026-07-16
Publish UpdateCatalog.jar + read datasources creds from env (#161)

### LINKED ISSUE #161: CI: publish UpdateCatalog.jar + accept datasources creds via env/file (#155/#156 follow-up)
Follow-up to #155/#156. rc-meta provisions a daily `UpdateCatalog` run (systemd timer, phase 33) to keep the `brouter-segments-4` checksum set current — but two things on the RouteConverter side make that operable:

**1. Publish `UpdateCatalog.jar` to the prerelease channel.**
`download-tools` already shades `target/UpdateCatalog.jar` (maven-shade). The host fetches tools from `https://releases.routeconverter.com/prerelease/` (that's how `RouteConverterCmdLine.jar` reaches the analyzer). Add `UpdateCatalog.jar` to whatever CI step publishes the cmdline jar so the timer host can `curl` it.

**2. Accept datasources credentials via env var or file (not just argv).**
`UpdateCatalog` only takes `--username/--password` on the command line, so when run as a systemd unit the password lands in `java`'s argv — visible in `ps` and the journal. Support `DATASOURCES_USERNAME`/`DATASOURCES_PASSWORD` env vars (or a `--password-file`) so the credential never hits the process command line.

Until both land, the rc-meta phase-33 timer stays provisioned-but-disabled (the jar fetch 404s; creds would be argv-exposed).

cpesch:
Follow-up to #155/#156; implements #161.

The rc-meta `update-catalog` timer runs `UpdateCatalog` daily to keep the `brouter-segments-4` checksum set current. Two changes make that operable:

**1. Publish `UpdateCatalog.jar` to the prerelease channel.**
`_build-linux-mac.yml` "Stage artefacts" now copies `download-tools/target/UpdateCatalog.jar` alongside `RouteConverterCmdLine.jar`, so it reaches `https://releases.routeconverter.com/prerelease/UpdateCatalog.jar` (where the timer host fetches it). Published **unsigned** — it's an internal ops tool, not a user download, so it stays out of the SignPath jar list.

**2. `UpdateCatalog` reads creds from the environment.**
`run()` now falls back to `DATASOURCES_SERVER` / `DATASOURCES_USERNAME` / `DATASOURCES_PASSWORD` env vars when the matching `--server/--username/--password` option is absent (small `optionOrEnv` helper). A systemd unit can inject them via `EnvironmentFile` so the password never lands in `java`'s argv (`ps` / journal). CLI args still take precedence — existing invocations are unchanged.

Verified: `./mvnw -pl download-tools -am package` builds `UpdateCatalog.jar` (Main-Class `slash.navigation.download.tools.UpdateCatalog`).

After merge, the rc-meta phase-33 unit can drop `--username/--password` from its `ExecStart` and rely on the injected env (small rc-meta follow-up).

Closes #161.

### DIFF
--- a/.github/workflows/_build-linux-mac.yml
+++ b/.github/workflows/_build-linux-mac.yml
@@ -102,6 +102,10 @@ jobs:
           cp RouteConverterMac/target/RouteConverterMac-x64-app.zip artefacts/
           cp RouteConverterMac/target/RouteConverterMac-aarch64-app.zip artefacts/
           cp RouteConverterCmdLine/target/RouteConverterCmdLine.* artefacts/
+          # Internal ops tool consumed by the rc-meta update-catalog timer
+          # (keeps the brouter-segments checksum set current, #155/#156/#161).
+          # Not user-facing, so it is published unsigned like a plain jar.
+          cp download-tools/target/UpdateCatalog.jar artefacts/
           cp TimeAlbumProLinux/target/TimeAlbumProLinux.* artefacts/
           cp TimeAlbumProMac/target/TimeAlbumProMac-x64-app.zip artefacts/
           cp TimeAlbumProMac/target/TimeAlbumProMac-aarch64-app.zip artefacts/
--- a/download-tools/src/main/java/slash/navigation/download/tools/UpdateCatalog.java
+++ b/download-tools/src/main/java/slash/navigation/download/tools/UpdateCatalog.java
@@ -368,11 +368,20 @@ public class UpdateCatalog extends BaseDownloadTool {
     private void run(String[] args) throws Exception {
         CommandLine line = parseCommandLine(args);
         setId(line.getOptionValue(ID_ARGUMENT));
-        setDataSourcesServer(line.getOptionValue(DATASOURCES_SERVER_ARGUMENT));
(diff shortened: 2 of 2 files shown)


cpesch (commented):

Small, well-scoped PR: publishes UpdateCatalog.jar as an unsigned internal ops artifact in the linux-mac build, and adds an env-var fallback (optionOrEnv) so datasources credentials can come from a systemd EnvironmentFile instead of process argv. Logic is correct, precedence (CLI over env) is preserved, and the change matches the PR description; no blockers found.

- **nit** [style] line 380 — New optionOrEnv fallback logic has no unit test; a small test asserting CLI-arg-wins-over-env and both-absent-returns-null would guard this credential-fallback path going forward.

---

cpesch:

**Reason:** feedback is an Inspector verdict marked comment-only/nit, not a maintainer change request — no actionable edit to make
**Confidence:** 0.3

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-11095` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-11095).

PR #178, opened 2026-07-19
ci(macos): gated Developer ID signing + notarisation pipeline

cpesch:
## What

Prepares the full macOS code-signing + notarisation pipeline so it works the moment an Apple Developer subscription lands — no core rewrite needed.

### New
- **`.github/workflows/_sign-mac.yml`** — reusable workflow, `macos-latest`. Signs each `RouteConverter.app` with a Hardened Runtime + JIT entitlements, then notarises + staples.
- **`.github/signing/entitlements.plist`** — JIT / unsigned-exec-mem / disable-library-validation / dyld-env entitlements for the bundled-JRE Swing app.

### Credential gate (runs green today)
Keys off the `MACOS_*` secrets:
- **present** → real Developer ID signing + `notarytool` + `stapler`
- **absent** → ad-hoc (`--sign -`), notarise/staple skipped

So the pipeline is exercised end-to-end **without an Apple subscription**; paste the 4 secrets later and it flips to real signing with zero YAML edits.

### Signing correctness
- Inner-out: JRE dylibs / `Contents/bin` / `jspawnhelper` first, bundle last. Never `--deep` for signing a notarised bundle.
- **Embedded-jar Mach-O:** `RouteConverterMac.jar` ships `sqlite-jdbc 3.53.2.0` (via `mapsforge-poi`/`mapsforge-mbtiles`) with `org/sqlite/native/Mac/{x86_64,aarch64}/libsqlitejdbc.dylib`. notarytool recurses into jars and would reject these as unsigned nested code — the workflow extract-sign-repacks them before sealing the bundle. bash-3.2-safe.

### Wiring
`sign-mac` job added to `prerelease.yml` + `release.yml`: consumes the unsigned linux-mac artefact, publishes signed app zips under a name **outside** the flatten glob, then publish overlays them over the unsigned ones deterministically (before the `-app.zip` → `.app.zip` rename in release).

## Secrets to add later
`MACOS_CERTIFICATE_BASE64`, `MACOS_CERTIFICATE_PASSWORD`, `MACOS_SIGN_IDENTITY`, `MACOS_NOTARY_PROFILE_B64` (base64 JSON `{appleId, teamId, password}`).

## Test now
Merge → next master push runs the ad-hoc dry-run (pack → sign → verify → overlay) green.

### DIFF
--- a/.github/signing/entitlements.plist
+++ b/.github/signing/entitlements.plist
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
+  "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!--
+  Hardened Runtime entitlements for the bundled-JRE RouteConverter.app.
+  - allow-jit / allow-unsigned-executable-memory: the JVM JITs bytecode into
+    executable memory pages at runtime.
+  - disable-library-validation: Java dlopen()s third-party dylibs (mapsforge,
+    JNI libs) that are not signed by our Team ID.
+  - allow-dyld-environment-variables: the launcher sets JAVA/DYLD env vars.
+  See rc-meta specs (macOS signing prep).
+-->
+<plist version="1.0"><dict>
+  <key>com.apple.security.cs.allow-jit</key><true/>
+  <key>com.apple.security.cs.allow-unsigned-executable-memory</key><true/>
+  <key>com.apple.security.cs.disable-library-validation</key><true/>
+  <key>com.apple.security.cs.allow-dyld-environment-variables</key><true/>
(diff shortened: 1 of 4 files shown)


cpesch (commented):

Well-structured, credential-gated macOS signing/notarisation pipeline that runs green without Apple secrets and flips to real signing when they land; logic (inner-out signing, embedded-jar dylib re-signing, artefact overlay ordering) checks out. One minor hardening nit on the throwaway keychain password.

### .github/workflows/_sign-mac.yml
- **minor** [security] line 78 — The temporary signing keychain's unlock password is a static literal rather than randomly generated per run; low risk since the keychain only exists for the ephemeral macos-latest runner's lifetime, but a generated value is better hygiene.
```suggestion
KC_PW="$(openssl rand -base64 24)"                # throwaway keychain password, unique per run
```

---

PR #182, opened 2026-07-20
ci(macos): sign .jnilib natives + guard unsigned nested Mach-O

cpesch:
Follow-up hardening on the macOS signing pipeline (#178).

## Changes
1. **Broaden embedded-native sign glob** `.dylib` → `.dylib|.jnilib`. JNA-style natives (`com/sun/jna/darwin-*/libjnidispatch.jnilib`) would otherwise ship unsigned inside a jar. Not a dep today, but a latent trap when one is added.
2. **Post-sign guard.** `codesign --deep` does not recurse into jars, so a Mach-O nested in a jar that the sign glob missed (new native dep, or a native under an unexpected extension) would only surface as a notarytool rejection. The guard magic-checks native-ish + extensionless jar entries with `file(1)`, skips ELF/PE, and fails the build on any unsigned Mach-O — closing the gap before notarisation. Ad-hoc signatures pass `codesign --verify`, so it also gates the dry-run.

Addresses forward-item #2 from the #178 wrap-up.

### DIFF
--- a/.github/workflows/_sign-mac.yml
+++ b/.github/workflows/_sign-mac.yml
@@ -148,20 +148,21 @@ jobs:
                     --sign "$SIGN_ID" --entitlements "$ENTITLEMENTS" {}
             fi
 
-            # 1b) Mach-O dylibs EMBEDDED inside the fat jar. notarytool scans
+            # 1b) Mach-O natives EMBEDDED inside the fat jar. notarytool scans
             # inside jars; sqlite-jdbc (via mapsforge-poi/mbtiles) ships
             # org/sqlite/native/Mac/{x86_64,aarch64}/libsqlitejdbc.dylib, which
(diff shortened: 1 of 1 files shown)


cpesch (commented):

Solid, well-reasoned hardening of the macOS embedded-native signing pipeline (broadens the sign glob to `.dylib`+`.jnilib` and adds a post-sign Mach-O magic-check guard); one minor edge case in the guard's candidate regex could let a root-level extensionless native slip past unnoticed.

### .github/workflows/_sign-mac.yml

- **minor** [correctness] line 193 — The guard's extensionless-candidate branch `/[^/.]+$` requires a preceding `/`, so a native binary stored at the jar root (no directory prefix, no extension) would not be picked up as a candidate and could bypass both the sign step and the guard.
```suggestion
grep -iE '\.(dylib|jnilib|so|dll)$|(^|/)[^/.]+$'
```

---

PR #194, opened 2026-07-20
fix(map): show world-map background under any displayed map, not only…

cpesch:
… Mapsforge

The offline edition skips the routeconverter/ directory when scanning maps (MapsforgeMapManager.scanMaps), so world.map is never a selectable map - it is only ever the background layer. When no offline map is installed, the sole available map is the online OpenStreetMap default (a Download map). But handleBackground attached the world.map background only when the displayed map was of type Mapsforge, so with OpenStreetMap selected and its tiles unreachable the map stayed gray with no base layer at all.

Treat world.map as what it is: a global, low-detail base layer that belongs underneath every displayed map type. shouldAttachBackground now attaches whenever the layer is built and a map is displayed. The background sits at layer index 0, so an opaque map on top hides it - no visible cost when the displayed map renders, and a usable world map instead of gray when it does not.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachment.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachment.java
@@ -24,6 +24,14 @@ package slash.navigation.mapview.mapsforge;
  * base layer, independent of the order in which the async background map
  * download (Path B) and the EDT map/theme layer-stack build (Path A) run.
  *
+ * The world map is a global, low-detail base layer: it belongs underneath every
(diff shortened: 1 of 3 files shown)


cpesch (commented):

Small, well-scoped fix that removes the Mapsforge-only gate from `BackgroundMapAttachment.shouldAttachBackground`, replacing it with a `layerReady && hasDisplayedMap` check so the world-map background can serve as a base layer under any displayed map type. The single call site in `MapsforgeMapView.handleBackground` and the unit tests are updated consistently with the new two-argument signature; the added Javadoc clearly explains the rationale (background at layer index 0, no visible cost when an opaque map renders on top).

No blockers found. No convention checks triggered.

---

PR #199, opened 2026-07-21
fix(formats): decode legacy ANSI KML/GPX as Windows-1252 when not valid UTF-8

### LINKED ISSUE #198: KML/GPX Garble formats: umlauts become mojibake for ANSI files under JRE 21 (3.4→3.5 regression)
# KML/GPX Garble formats: umlauts turn to mojibake for ANSI files under JRE 21 (3.4→3.5 regression)

## Summary
Users with hand-written KML POI files saved in Windows ANSI (Windows-1252), **without** an XML encoding declaration and **without** a BOM, see umlauts/diacritics as garbage (`Gr�n Stra�e`) starting with 3.5. Up to 3.4 the same files read correctly. Reported by a long-time user.

## Root cause (reproduced and proven)
Such a file is not valid UTF-8, so the **normal** `Kml22Format` path (`Unmarshaller.unmarshal(InputStream)` → XML parser assumes UTF-8) throws. RouteConverter then falls back to the salvage format **`GarbleKml22Format`**, whose `read()` does:

```java
try (InputStreamReader reader = new InputStreamReader(source)) {   // NO explicit charset
    KmlType kmlType = unmarshal22(reader);
```
`navigation-formats/.../kml/GarbleKml22Format.java:50`

`new InputStreamReader(source)` with no charset uses the **JVM default charset**. **JEP 400** changed that default to **UTF-8 in JDK 18+**; on JDK ≤17 it was the platform default (Windows-1252 on a German Windows). RouteConverter 3.4→3.5 bumped the bundled JRE **17 → 21** (`b98227f8a`), flipping the default. Result: the ANSI bytes are decoded as UTF-8, high bytes become U+FFFD → mojibake. In 3.4 (JRE 17) the same reader used Windows-1252 and read correctly.

**Proof (same jars, JDK 21, only the default charset varied):**
```
-Dfile.encoding=UTF-8         → name=[Gr�n � � � Stra�e]   (U+FFFD, matches the report)
-Dfile.encoding=windows-1252  → name=[Grün Ö Ä ß Straße]   (correct — the JRE-17 behavior)
```
Hypotheses ruled out along the way: the javax→jakarta migration (was at 3.1~25, before 3.4); a StAX-level difference between JDK 17 and 21 (identical). The decisive factor is `InputStreamReader`'s default charset, not XML autodetection.

## Fix (implemented + verified)
The charset-less readers must not depend on the JVM default. Naming a fixed charset would only trade one platform bias for another (hard-coding Windows-1252 breaks genuine UTF-8 on macOS/Linux; `native.encoding` re-inherits the old locale lottery and isn't reproducible in tests). Instead, decode **deterministically**: try strict UTF-8, and fall back to Windows-1252 only when the bytes are not valid UTF-8. This reads real UTF-8 correctly *and* salvages legacy Western-ANSI files correctly on every OS, independent of locale.

Shared helper in `common` `InputOutput`:
```java
public static Reader readAsReaderPreferringUtf8(InputStream input) throws IOException {
    byte[] bytes = readBytes(input);
    return new StringReader(new String(bytes, isValidUtf8(bytes) ? UTF_8 : WINDOWS_1252));
}
```
All charset-less XML salvage readers route through it (6): `kml/GarbleKml22Format`, `kml/GarbleKml21Format`, `kml/GarbleKml22BetaFormat`, `gpx/GarbleGpx10Format`, `gpx/GarbleGpx11Format`, `viamichelin/ViaMichelinFormat`. (`kml/GarbleKml21LittleEndianFormat` already passes `UTF_16LE`; `csv/CsvFormat` and `base/TextNavigationFormat` already pass an explicit charset.)

Verified on a default-UTF-8 JVM (JDK 21): the ANSI fixture now reads `Grün Ö Ä ß Straße` (no replacement char), real UTF-8 stays correct, and a `windows-1252`-declared file stays correct. Unit tests added in `InputOutputTest` cover the UTF-8 branch, the Windows-1252 fallback branch, and ASCII (`common` build green, 11/11).

Non-Western ANSI (e.g. Greek Cp1253) is still not recovered — those never worked cross-platform and are out of scope for a Western-umlaut regression.

## Workaround (already given to reporter)
Add to the first line of the KML file: `<?xml version="1.0" encoding="windows-1252"?>` (or `ISO-8859-1`). Then the file parses via the normal path and umlauts read correctly; no UTF-8 conversion needed. Confirmed against the current build.

## Acceptance
- ANSI KML with umlauts, no prolog/BOM → reads correctly on JRE 21 (regression fixed).
- UTF-8 KML (with/without declaration) → unchanged.
- KML declaring `encoding="windows-1252"`/`ISO-8859-1` → honored (already works).
- Test fixtures for all three cases in `navigation-formats` (the three used to verify are in the issue thread).

cpesch:
Fixes #198

## Problem
Hand-written KML/GPX POI files saved in Windows ANSI (Windows-1252) with no XML encoding declaration and no BOM show umlauts as `Gr�n Stra�e` since 3.5; they read correctly up to 3.4.

## Root cause (reproduced)
Such a file is not valid UTF-8, so the normal `Kml22Format` parse (`unmarshal(InputStream)` → XML parser assumes UTF-8) throws, and RouteConverter falls back to the salvage format `GarbleKml22Format`. That reader used `new InputStreamReader(source)` **with no explicit charset** → the JVM default. **JEP 400** made the JVM default UTF-8 in JDK 18+, and 3.4→3.5 bumped the bundled JRE **17 → 21**, so the ANSI bytes now decode as UTF-8 → U+FFFD. Proven: same jars, JDK 21, only the default charset varied — `file.encoding=UTF-8` → mojibake, `windows-1252` → correct.

## Fix
Decode **deterministically**, independent of the platform default charset: try strict UTF-8, fall back to Windows-1252 only when the bytes aren't valid UTF-8. A fixed charset would just trade one platform bias for another (hard-coded 1252 breaks genuine UTF-8 on macOS/Linux; `native.encoding` re-inherits the old locale lottery and isn't test-reproducible). This reads real UTF-8 *and* salvages legacy Western-ANSI files identically on every OS.

New shared helper `InputOutput.readAsReaderPreferringUtf8(InputStream)`; all charset-less XML salvage readers route through it: `GarbleKml22Format`, `GarbleKml21Format`, `GarbleKml22BetaFormat`, `GarbleGpx10Format`, `GarbleGpx11Format`, `ViaMichelinFormat`. (`GarbleKml21LittleEndianFormat` already uses `UTF_16LE`; `CsvFormat`/`TextNavigationFormat` already pass an explicit charset.)

## Verification
- On a default-UTF-8 JVM (JDK 21): ANSI KML now reads `Grün Ö Ä ß Straße` (no replacement char); genuine UTF-8 and `windows-1252`-declared files stay correct.
- Unit tests added in `InputOutputTest` for the UTF-8 branch, the Windows-1252 fallback branch, and ASCII — `common` build green (11/11).

## Notes
- User-facing workaround for existing files (no upgrade needed): add `<?xml version="1.0" encoding="windows-1252"?>` as the first line.
- Out of scope: non-Western ANSI code pages (e.g. Greek Cp1253) — never worked cross-platform.

### DIFF
--- a/common/src/main/java/slash/common/io/InputOutput.java
+++ b/common/src/main/java/slash/common/io/InputOutput.java
@@ -21,8 +21,13 @@
 package slash.common.io;
 
 import java.io.*;
+import java.nio.ByteBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
 
+import static java.nio.charset.CodingErrorAction.REPORT;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.apache.commons.io.IOUtils.copyLarge;
 
@@ -55,5 +60,27 @@ public class InputOutput {
 
     public static String readFileToString(File file) throws IOException {
         return new String(readBytes(new FileInputStream(file)), UTF_8);
+    }
+
+    private static final Charset WINDOWS_1252 = Charset.forName("windows-1252");
+
+    /**
+     * Returns a {@link Reader} for legacy/garbled XML that carries no encoding declaration: decodes as
+     * UTF-8 when the bytes are valid UTF-8, otherwise falls back to Windows-1252 (the common legacy code
+     * page). Deterministic and independent of the platform default charset, which changed to UTF-8 in
+     * JDK 18 (JEP 400) and previously made this depend on the host locale.
+     */
+    public static Reader readAsReaderPreferringUtf8(InputStream input) throws IOException {
+        byte[] bytes = readBytes(input);
+        return new StringReader(new String(bytes, isValidUtf8(bytes) ? UTF_8 : WINDOWS_1252));
+    }
+
+    private static boolean isValidUtf8(byte[] bytes) {
+        try {
+            UTF_8.newDecoder().onMalformedInput(REPORT).onUnmappableCharacter(REPORT).decode(ByteBuffer.wrap(bytes));
+            return true;
+        } catch (CharacterCodingException e) {
+            return false;
+        }
     }
 }
--- a/common/src/main/java/slash/common/io/InputOutput.java
+++ b/common/src/main/java/slash/common/io/InputOutput.java
@@ -24,8 +24,6 @@ import java.io.*;
 import java.nio.ByteBuffer;
 import java.nio.charset.CharacterCodingException;
 import java.nio.charset.Charset;
-import java.nio.charset.CodingErrorAction;
-import java.nio.charset.StandardCharsets;
 
 import static java.nio.charset.CodingErrorAction.REPORT;
 import static java.nio.charset.StandardCharsets.UTF_8;
--- a/common/src/test/java/slash/common/io/InputOutputTest.java
+++ b/common/src/test/java/slash/common/io/InputOutputTest.java
@@ -25,6 +25,7 @@ import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
 import java.io.*;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 
 import static org.junit.Assert.*;
@@ -106,6 +107,34 @@ public class InputOutputTest {
         }
         String result = InputOutput.readFileToString(file);
         assertEquals(content, result);
+    }
+
+    private static String toString(Reader reader) throws IOException {
+        StringWriter writer = new StringWriter();
+        InputOutput.copyAndClose(reader, writer);
+        return writer.toString();
+    }
+
+    @Test
+    public void testReadAsReaderPreferringUtf8DecodesUtf8() throws IOException {
+        String text = "Grün Ö Ä ß Straße";
+        Reader reader = InputOutput.readAsReaderPreferringUtf8(new ByteArrayInputStream(text.getBytes(StandardCharsets.UTF_8)));
+        assertEquals(text, toString(reader));
+    }
+
+    @Test
+    public void testReadAsReaderPreferringUtf8FallsBackToWindows1252() throws IOException {
+        String text = "Grün Ö Ä ß Straße";
+        byte[] ansi = text.getBytes(Charset.forName("windows-1252")); // not valid UTF-8
+        Reader reader = InputOutput.readAsReaderPreferringUtf8(new ByteArrayInputStream(ansi));
+        assertEquals(text, toString(reader));
+    }
+
+    @Test
+    public void testReadAsReaderPreferringUtf8PlainAscii() throws IOException {
+        String text = "plain ascii 123";
+        Reader reader = InputOutput.readAsReaderPreferringUtf8(new ByteArrayInputStream(text.getBytes(StandardCharsets.US_ASCII)));
+        assertEquals(text, toString(reader));
     }
 }
 
--- a/navigation-formats/src/main/java/slash/navigation/csv/CsvFormat.java
+++ b/navigation-formats/src/main/java/slash/navigation/csv/CsvFormat.java
@@ -31,11 +31,11 @@ import slash.navigation.common.NavigationPosition;
 
 import java.io.*;
 import java.util.*;
-import java.util.logging.Logger;
 
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static java.util.Arrays.sort;
+import static slash.common.io.InputOutput.readAsReaderPreferringUtf8;
 import static slash.common.io.Transfer.*;
 
 /**
@@ -45,7 +45,6 @@ import static slash.common.io.Transfer.*;
  */
 
 public abstract class CsvFormat extends BaseNavigationFormat<CsvRoute> {
-    private static final Logger log = Logger.getLogger(CsvFormat.class.getName());
 
     public String getExtension() {
         return ".csv";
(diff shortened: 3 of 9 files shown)


cpesch (commented):

Well-scoped, well-tested fix: a shared UTF-8-with-Windows-1252-fallback reader replaces bare `InputStreamReader` in six charset-less salvage/Garble format readers, addressing the JEP 400 default-charset regression cleanly. Only one minor nit found; no blockers, no convention-check hits.

### common/src/main/java/slash/common/io/InputOutput.java
- **nit** [style] line 27 — `CodingErrorAction` is imported directly but never referenced by its qualified name — only the static import of `REPORT` is used, making this import dead.
```suggestion
import java.nio.charset.StandardCharsets;
```

---

cpesch:
🤖 Patcher pushed a fix to `fix/garble-charset-utf8-fallback`.

cpesch:

Solid, well-tested fix for the JEP-400 charset regression with a clear root-cause writeup, but the refactor changes stream-closing semantics: the new `readAsReaderPreferringUtf8()` helper buffers the InputStream into a String/StringReader, so the try-with-resources in all six call sites no longer closes the original source stream the way the replaced `InputStreamReader` did — worth confirming/fixing before merge.

### common/src/main/java/slash/common/io/InputOutput.java

- **major** [correctness] line 72 — `readAsReaderPreferringUtf8()` reads the InputStream fully via `readBytes(input)` and returns a `StringReader` wrapping the resulting String. Unlike the replaced `new InputStreamReader(source)` (whose `close()` cascaded to `source.close()`), the try-with-resources in each of the six call sites (`GarbleKml21/22/22Beta`, `GarbleGpx10/11`, `ViaMichelinFormat`) now only closes the `StringReader`, which is a no-op — the original `source` stream is left unclosed unless `InputOutput.readBytes()` closes its input parameter internally. Please confirm `readBytes()` closes the stream, or close it explicitly here, to avoid a handle leak on every garbled-format parse attempt.
```suggestion
    public static Reader readAsReaderPreferringUtf8(InputStream input) throws IOException {
        byte[] bytes;
        try (InputStream in = input) {
            bytes = readBytes(in);
        }
        return new StringReader(new String(bytes, isValidUtf8(bytes) ? UTF_8 : WINDOWS_1252));
    }
```
- **nit** [style] line 72 — This now fully materializes the file into a `byte[]` and decodes it twice (once to validate UTF-8, once to build the String) instead of streaming through a Reader. Fine for hand-written POI files but worth a one-line comment noting the trade-off if very large garbled inputs are ever expected.

---

cpesch (commented):
## Verdict: approve · confidence 0.92

Deterministic UTF-8-then-Windows-1252 fallback for charset-less XML salvage readers, centralized in the shared `readAsReaderPreferringUtf8` helper and covered by unit tests (UTF-8, Windows-1252 fallback, ASCII). This is the second review pass (prior patch round 1); the fix is sound and merge-ready — no blocker or major findings remain.

---

cpesch:
Folded in an additional commit extending the same helper to CSV:

- `refactor(csv)`: `CsvFormat` now decodes via `InputOutput.readAsReaderPreferringUtf8` instead of its bespoke UTF-8→ISO-8859-1 two-pass (`mark`/`available`/`reset` + the `isIsoLatin1ButReadWithUtf8` garbage heuristic). Same result for UTF-8 and legacy ANSI CSV, deterministically; the per-format garbage check is now redundant and removed. (`Transfer.isIsoLatin1ButReadWithUtf8` stays — TomTom still references it.)

Also investigated **TomTom .itn** and **TextNavigationFormat** as further candidates, and deliberately left both alone:
- **TextNavigationFormat**: its `read(InputStream)` default (ISO-8859-1) is shared by the whole line-based family; flipping it would change reading *and* format detection broadly. Out of scope here.
- **TomTom .itn**: already reads both encodings correctly today — the TomTom5 (ISO) / TomTom9.5 (UTF-8, garble-reject) pair handles it. Verified: a UTF-8 .itn reads as "Tom Tom 9.5" and an ANSI .itn as "Tom Tom 5", both with intact umlauts. Applying the helper there would add no fault-tolerance and could flip the detected format label (hence the write-back encoding), so it was not changed.

Tests: `common` 11/11 (incl. new `InputOutputTest` cases), `navigation-formats` CSV + TomTom incl. read/write round-trips all green.

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: Working tree is clean and my verification confirms the maintainer's requested change is already fully implemented on this branch — I'm making no edits.

No changes needed: CSV refactor already committed and verified correct

The feedback describes work that is already present in this checkout's history:

- `6aaeaf967` added the shared `InputOutput.readAsReaderPreferringUtf8(InputStream)` helper and routed the six charset-less XML "Garble" readers (`GarbleKml22Format`, `GarbleKml21Format`, `GarbleKml22BetaFormat`, `GarbleGpx10Format`, `GarbleGpx11Format`, `ViaMichelinFormat`) through it.
- `9ffb11dfe` (an earlier, mis-labeled automated commit) removed two now-unused imports (`CodingErrorAction`, `StandardCharsets`) that the first commit had left dangling in `InputOutput.java`.
- `387b5ffa2` is exactly the "additional commit extending the same helper to CSV" the feedback asks for: `CsvFormat.java` now decodes via `readAsReaderPreferringUtf8`, and the old UTF-8→ISO-8859-1 two-pass plus the `containsGarbage`/`isIsoLatin1ButReadWithUtf8` per-row heuristic were removed.

I re-verified the current working tree (not just the commit messages) via a dedicated read-only review:

- `InputOutput.java` — no unused imports/constants, `readAsReaderPreferringUtf8`/`isValidUtf8` are correct and self-contained.
- All 6 Garble/ViaMichelin readers correctly import and call the helper, no leftover default-charset `InputStreamReader` usage.
- `CsvFormat.java` is internally consistent — no dangling calls to the removed `containsGarbage` or old `read(InputStream, String, ParserContext)` signature.
- `Transfer.isIsoLatin1ButReadWithUtf8`, `UTF8_ENCODING`, `ISO_LATIN1_ENCODING` are untouched and still used by the TomTom formats, matching the feedback's explicit "TomTom stays as-is" decision.
- `InputOutputTest.java` has the three described unit tests (UTF-8, Windows-1252 fallback, ASCII).
- Repo-wide search found zero remaining bare (`new InputStreamReader(source)`, no charset) call sites and
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-12833` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-12833).

PR #203, opened 2026-07-22
Fix two Maps dialog crashes (double-open + null bounding box)

cpesch:
Fixes two independent crashes in the Maps dialog, both reported by a user (Edgar Kraus) on macOS with RouteConverter 3.5-SNAPSHOT-298.

## 1. `IllegalArgumentException: Action '…' for 'display-online-map' already registered`

`SingletonDialogAction` — despite its name and Javadoc ("Show a dialog at most once") — built a new dialog on every invocation. `MapsDialog` (and the themes/downloads dialogs) register their actions in the constructor and unregister them in `close()`. Opening a second dialog while the first was still showing therefore re-registered the same action key and threw, crashing `ShowMapsAction`.

Fix: cache the dialog instance; if it is still showing, bring it to front instead of constructing a duplicate. Applies to all `SingletonDialogAction` subclasses (Maps, Themes, Downloads, Options, About, …).

### Lineage — why this isn't a plain revert

`SingletonDialogAction` originally cached the dialog (commit `e9d542def`, "introduce SingletonDialogAction … to prevent them being opened more than once"). That cache was removed in `776a07648` (2024-07-03), bundled into an unrelated logout commit: after the change `run()` built a fresh dialog on every open.

The original cache had the opposite defect — it never reset the field, so it reused the **same** dialog instance forever and showed **stale** content on reopen (models are bound once at construction). Removing it fixed staleness but reintroduced the double-open crash the class exists to prevent.

This fix keeps both properties: it rebuilds the dialog when it is **not** currently showing (fresh content, no staleness), and reuses / fronts it only while it **is** showing (no duplicate registration).

## 2. `NullPointerException` when selecting a downloadable map

`GraphDescriptor.matchesBoundingBox` null-checked the graph's own bounding box but passed the map's bounding box straight into `BoundingBox.contains(BoundingBox)`, which dereferences `other.northEast()`. Selecting a downloadable map whose descriptor has no bounding box crashed via `MapsDialog.updateLabel` → `GraphHopper.calculateRemainingDownloadSize` → `DownloadableFinder`.

Fix: skip the bounding-box match when the map descriptor has no bounding box.

Both modules compile (`common-gui`, `graphhopper`).

### DIFF
--- a/common-gui/src/main/java/slash/navigation/gui/actions/SingletonDialogAction.java
+++ b/common-gui/src/main/java/slash/navigation/gui/actions/SingletonDialogAction.java
@@ -29,11 +29,16 @@ import slash.navigation.gui.SimpleDialog;
  */
 
 public abstract class SingletonDialogAction extends FrameAction {
+    private SimpleDialog dialog;
 
     protected abstract SimpleDialog createDialog();
 
     public void run() {
-        SimpleDialog dialog = createDialog();
+        if (dialog != null && dialog.isShowing()) {
+            dialog.toFront();
+            return;
+        }
+        dialog = createDialog();
         dialog.showWithPreferences();
         dialog.toFront();
         dialog.setVisible(true);
--- a/common-gui/src/test/java/slash/navigation/gui/actions/SingletonDialogActionTest.java
+++ b/common-gui/src/test/java/slash/navigation/gui/actions/SingletonDialogActionTest.java
@@ -1,0 +1,70 @@
+/*
+    This file is part of RouteConverter.
+
+    RouteConverter is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    RouteConverter is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with RouteConverter; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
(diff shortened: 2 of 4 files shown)


cpesch (commented):

Two small, well-targeted crash fixes (singleton dialog re-open guard, null map bounding box in `matchesBoundingBox`); logic matches the described bugs. Flagging one speculative edge case and a test-coverage nit — neither blocks merging.

### common-gui/src/main/java/slash/navigation/gui/actions/SingletonDialogAction.java

- **minor** [correctness] line 37 — The guard relies on `dialog.isShowing()` to decide whether to reuse or recreate the dialog, but the action-registration crash this PR fixes is a constructor/`close()` pairing — if a subclass's dialog can be dismissed via the window manager's close box (or `dispose()`) without going through the `close()` method that unregisters the action, `isShowing()` will correctly return false, `createDialog()` will run again, and the original registration crash could resurface for that subclass. Worth confirming that all `SingletonDialogAction` subclasses (Maps, Themes, Downloads, Options, About) only ever hide/dispose via their own `close()`.

### graphhopper/src/main/java/slash/navigation/graphhopper/GraphDescriptor.java

- **nit** [correctness] line 47 — `matchesBoundingBox` is a plain, easily-testable method (no GUI dependency) — a small unit test asserting the null-map-bounding-box case returns false would lock in this regression fix.

---

cpesch:
🤖 Patcher pushed a fix to `fix/maps-dialog-singleton-double-register`.

cpesch (commented):

Second-round review: the singleton-dialog reuse guard and the bounding-box null-check both correctly address the reported crashes, and the added GraphManagerTest regression test exercises the second fix. No blockers found; closing out the polish round.

---

cpesch (commented):

Second-round review: `SingletonDialogAction` now caches the dialog and only rebuilds when it is not currently showing (fixes the double-registration crash while avoiding stale content), and `GraphDescriptor` null-checks the map's bounding box before calling `contains()`, fixing the NPE on map selection. Both fixes are covered by new unit tests exercising the exact crash scenarios; no remaining correctness, security, or convention issues found.

---

Issue #204, opened 2026-07-22
[forum #4139] macOS "Open with" drops the file on a cold launch (odoc not handled)


cpesch:
## Bug: macOS "Open with" drops the right-clicked file on a cold launch

**Source:** https://forum.routeconverter.com/thread-4139-post-23143.html#pid23143. Request (verbatim): "If I right-click on a .CSV or .gpx file and select \"Open with\" and then choose RouteConverter - the RouteConverter program launches but does not open the file that I initially right-clicked on. … Once RouteConverter has launched, I can then go into File > Open... and select the .CSV or .gpx file and the route file opens as expected. The \"right-click > Open with\" process always worked on previous versions, however this version (3.5-SNAPSHOT-103) doesn't seem to be working." Goal: on macOS, a Finder "Open with → RouteConverter" (or `open -a` / dock-drop) on a cold, not-yet-running app must load the chosen file — as it did before 3.5.

Reproduced locally on 2026-07-22 (x64 `.app`, Temurin 21): cold `open -a RouteConverter.app granada.gpx` → window opens, file NOT loaded; `File > Open` on the same file then works. Confirms the report.

This spec is fully resolved (grilled 2026-07-22); implement without further questions.

### Root cause
macOS "Open with" on a **cold** launch does not pass the file as `argv`; it launches the `.app`, then delivers the path via a `kAEOpenDocuments` (`odoc`) Apple Event.

Commit `c4f0c5001` (2026-06-13, "Mac apps: replace x86_64-only native launcher…") replaced the native `libapplauncher` Cocoa stub with a shell launcher that `exec`s `bin/java -jar` (`RouteConverterMac/src/main/app-resources/MacOS/RouteConverter`). The native stub installed the `odoc` Apple Event handler at process start and replayed the file to the JVM. The shell launcher does not; the JVM's AppKit/`eawt` Apple Event handler installs only seconds into JVM boot, and on a cold launch the `odoc` event is delivered (and dropped) before then.

Compounded by **late handler registration**: `java.awt.Desktop.setOpenFileHandler` is wired in `BaseRouteConverter.openFrame()` — i.e. after `initializeServices/Actions/Datasources` — widening the drop window. The handler code in `helpers/ApplicationMenu.java` (`openFiles`) is otherwise correct. RC already handles files passed on the **command line** via `BaseRouteConverter.parseInitialArgs()` (called by `Application.launch` right after `startup()`), but `odoc` never reaches that path.

### Scope
[...]

PR #205, opened 2026-07-22
[forum #4139] macOS "Open with" drops the file on a cold launch (odoc not handled)

cpesch:
## Summary

Implemented the fix for issue #204 (macOS "Open with" dropping files on cold launch) exactly per the locked spec.

**Root cause:** on a cold launch, macOS delivers the "Open with" file via a `kAEOpenDocuments` (`odoc`) Apple Event almost immediately, but the app only registered its `Desktop.setOpenFileHandler` deep into `openFrame()` — well after AppKit could have already fired and dropped the event.

### Changes

1. **`common-navigation/.../PendingOpenUrls.java`** (new) — pure, testable seam. `synchronized boolean addOrDefer(List<URL>)` buffers URLs until the UI is ready (returns `true` while buffering, `false` once released). `synchronized List<URL> release(List<URL> initialArgs)` marks the instance ready and returns `argv` URLs first, then buffered URLs, deduped by URL with order preserved.

2. **`common-navigation/.../PendingOpenUrlsTest.java`** (new) — covers all 5 scenarios from the spec: buffer-then-release, argv+buffered merge/dedup, empty release, `addOrDefer` after release, and multiple buffered calls accumulating in order.

3. **`ApplicationMenu.java`** — now takes a `PendingOpenUrls` in its constructor. `openFiles`/`openUri` defer to `pendingOpenUrls.addOrDefer(urls)`; if not buffered (warm launch), they open the URLs via `invokeLater` on the EDT.

4. **`BaseRouteConverter.java`** — owns one `PendingOpenUrls` instance. `new ApplicationMenu(pendingOpenUrls).addApplicationMenuItems()` is now the **first statement** in `startup()` (before `initializeDiagnostics()`), so the AppKit `odoc` handler arms as early as possible. The old registration in `openFrame()` was removed (registered exactly once now). `parseInitialArgs` now calls `pendingOpenUrls.release(toUrls(args))` and opens the merged list, or calls `newFile()` if empty — preserving existing no-file behavior while closing the race.

Threading is safe: all buffering/release is `synchronized`, and both the warm-path open and the flushed open run via `invokeLater` on the EDT.

### Note on verification

This sandboxed environment's Bash tool blocks `java`/build-tool invocations (approval required but unavailable), so I could not run `./mvnw test` or the manual macOS cold-launch repro myself. I instead manually traced `Application.launch()` → `startup()` → `parseInitialArgs()` to confirm both run on the EDT in the expected order, and reviewed the new code line-by-line against the 5 required test scenarios. The workflow's CI build should run `PendingOpenUrlsTest` and the full `mvn verify`; the macOS cold-launch manual check from the acceptance criteria still needs a human/CI runner with a real `.app` build.

Source issue: cpesch/RouteConverter#204

---
**Builder stats** · model `sonnet` · confidence `0` · 3m 13s across 1 round(s) · 4 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/13096)

### DIFF
--- a/common-navigation/src/main/java/slash/navigation/common/PendingOpenUrls.java
+++ b/common-navigation/src/main/java/slash/navigation/common/PendingOpenUrls.java
@@ -1,0 +1,65 @@
+/*
+    This file is part of BaseRouteConverter.
+
+    BaseRouteConverter is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    BaseRouteConverter is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with BaseRouteConverter; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+    Copyright (C) 2007 Christian Pesch. All Rights Reserved.
+*/
+
+package slash.navigation.common;
+
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+
+/**
+ * Buffers files/URLs delivered by the operating system (e.g. a macOS {@code odoc}
+ * Apple Event on a cold launch) before the UI is ready to open them, then merges
+ * them with the URLs passed as command line arguments once the UI becomes ready.
+ *
+ * @author Christian Pesch
+ */
+
+public class PendingOpenUrls {
+    private final List<URL> buffered = new ArrayList<>();
+    private boolean released = false;
+
+    /**
+     * @return {@code true} if the urls were buffered since the UI is not yet ready;
+     * {@code false} if the caller should open them immediately since this instance
+     * has already been {@link #release released}
+     */
+    public synchronized boolean addOrDefer(List<URL> urls) {
+        if (released)
+            return false;
+        buffered.addAll(urls);
+        return true;
+    }
+
+    /**
+     * Marks this instance as ready and returns the merged list of URLs to open:
+     * the given {@code initialArgs} first, then any buffered URLs, deduped by URL
+     * with order preserved. After this call {@link #addOrDefer} always returns
+     * {@code false}.
+     */
+    public synchronized List<URL> release(List<URL> initialArgs) {
+        released = true;
+        LinkedHashSet<URL> merged = new LinkedHashSet<>(initialArgs);
+        merged.addAll(buffered);
+        buffered.clear();
+        return new ArrayList<>(merged);
+    }
+}
--- a/common-navigation/src/main/java/slash/navigation/common/PendingOpenUrls.java
+++ b/common-navigation/src/main/java/slash/navigation/common/PendingOpenUrls.java
@@ -57,6 +57,10 @@ public class PendingOpenUrls {
      */
     public synchronized List<URL> release(List<URL> initialArgs) {
         released = true;
+        // Note: URL#equals()/hashCode() perform host resolution (a blocking DNS call) for
+        // URLs with a non-empty host. Harmless here since only file:// URLs with an empty
+        // host flow through this method; if this class is ever used for remote URLs,
+        // dedupe by URI (or String) instead to avoid a network call on the EDT.
         LinkedHashSet<URL> merged = new LinkedHashSet<>(initialArgs);
         merged.addAll(buffered);
         buffered.clear();
--- a/common-navigation/src/test/java/slash/navigation/common/PendingOpenUrlsTest.java
+++ b/common-navigation/src/test/java/slash/navigation/common/PendingOpenUrlsTest.java
@@ -1,0 +1,74 @@
+package slash.navigation.common;
+
+import org.junit.Test;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.List;
+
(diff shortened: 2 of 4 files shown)


cpesch (commented):

Solid, well-tested fix for the cold-launch odoc race: `PendingOpenUrls` buffers Apple Events until `parseInitialArgs` releases them, and `ApplicationMenu` registration moved to the top of `startup()` with only one registration site now. No blockers found; two minor design/verification notes below.

### common-navigation/src/main/java/slash/navigation/common/PendingOpenUrls.java

- **minor** [style] line 61 — Dedup uses a `LinkedHashSet<URL>`, and `java.net.URL.equals()`/`hashCode()` perform host resolution (potential blocking network/DNS call) for URLs with a non-empty host; harmless today since only `file://` URLs with empty host flow through here, but it's a well-known Java pitfall worth a comment or a switch to `URI` if this path is ever extended to remote URLs.

### route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ApplicationMenu.java

- **minor** [correctness] line 55 — `ApplicationMenu`'s constructor signature changed from no-arg to requiring `PendingOpenUrls`; the PR description notes the sandbox couldn't run mvn/build, so this hasn't been compile-verified against any other call sites (tests, other modules) that may still construct `ApplicationMenu()` with no args.

---

cpesch:
🤖 Patcher pushed a fix to `agent/builder/204-forum-4139-macos-open-with-drops-the-fil`.

cpesch (commented):

Second-round review: the fix cleanly closes the cold-launch odoc race via a synchronized `PendingOpenUrls` buffer, moves menu/handler registration to the top of `startup()`, and adds EDT dispatch (`invokeLater`) for Apple Event callbacks that previously invoked Swing code off-thread. Tests cover the five spec scenarios (buffer-then-release, argv+buffered merge/dedup, empty release, post-release `addOrDefer`, ordered accumulation). Only a self-documented, already-mitigated caveat remains (`URL#equals()` performing host resolution in the dedup path) — minor, doesn't block merge.

---

cpesch:
## Empirical verdict: this approach does not fix the bug — closing

Built this branch into a faithful macOS x64 `.app` (bundled JRE + shell launcher, identical to the release bundle) and tested the actual cold-launch "Open with" path in a real GUI session.

**Result: the file is still not loaded.**

Evidence (RC log, `-Djava.io.tmpdir` pinned so the run is captured), repeated across launches:

```
INFORMATION: Started RouteConverter 3.5-SNAPSHOT-local … for Mac …
INFORMATION: Processing initial arguments: []
INFORMATION: Processing urls: []
```

`openFiles` never fires; the pending buffer is empty because there is nothing to buffer.

Positive control (same app, same session):
- `open -n -a App --args ~/Desktop/granada.gpx` → **loads** (file arrives via argv → `parseInitialArgs` → `openUrls`). App / launcher / `openUrls` / `PendingOpenUrls` pipeline is correct; unit test `PendingOpenUrlsTest` is green (5/5) and the full build compiles.
- `open -a App ~/Desktop/granada.gpx` (the real odoc path) → **does not load**.

**Why Option 1 can't work:** the `kAEOpenDocuments` (odoc) Apple Event never reaches the JVM when the `.app` is launched via the shell `exec java -jar` launcher on a cold start. The removed native `libapplauncher` installed the AE handler in native code milliseconds after fork and caught the event; the JVM's AppKit/`eawt` handler installs seconds into startup, after macOS has already delivered (and dropped) the event. Registering `Desktop.setOpenFileHandler` earlier in `startup()` does not help — there is no event at the Java layer to catch or buffer.

**Real fix = Option 2:** a native launcher that handles odoc (jpackage-built launcher, or a small universal-binary stub), tracked in a separate spec. Closing this PR; the buffer/merge logic here (`PendingOpenUrls`) is sound and can be reused if a native launcher forwards odoc, but it is not sufficient on its own.

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: No safe fix is possible within this PR's scope — closing without code changes

I reviewed the maintainer's feedback (`cpesch`, maintainer) on PR #205 and made **no edits** to the working tree.

**Why no edit is appropriate here:**

1. **The maintainer performed empirical testing**, not code review. They built a real macOS `.app` bundle and confirmed via log evidence that the `PendingOpenUrls`/`ApplicationMenu` buffering approach in this PR does not fix the underlying bug: on a cold "Open with" launch, `kAEOpenDocuments` never reaches the JVM at all (`Processing urls: []`), so there is nothing at the Java layer for `PendingOpenUrls` to buffer or `Desktop.setOpenFileHandler` to catch — no matter how early it's registered in `startup()`.

2. **The maintainer explicitly states the verdict is architectural, not a bug in this diff**: the real fix ("Option 2") requires a *native* launcher (jpackage-built launcher or a universal-binary stub) that installs an AppleEvent handler in native code before/at process fork — restoring what the removed `libapplauncher` used to do. That is a fundamentally different mechanism from anything achievable by editing Java source in `common-navigation`/`ApplicationMenu`/`BaseRouteConverter`.

3. **The maintainer is closing PR #205** and says the native-launcher fix should be "tracked in a separate spec" — i.e., out of scope for this PR/branch. There is no incremental patch to *this* branch's Java code that would make the odoc cold-launch case work; attempting one would just be more code in the same dead-end direction the maintainer already disproved.

4. Per this task's own instructions ("If no safe fix is possible, make NO edits and explain why"), and per AGENTS.md's guidance to stop for "real product/compatibility/architecture decisions" — a native launcher redesign is exactly such a decision, requiring its own spec and PR, not a reactive patch inside this one.

**Salvageable part** (noted for whoever opens the follow-up native-launcher spe
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-13117` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-13117).

PR #216, opened 2026-07-23
fix(mac): build + host universal launcher stub on macOS runner (unblocks Mac release build)

cpesch:
Follow-up to #206/#208. The launcher stub is correct, but the `.app` is assembled in `_build-linux-mac.yml` on **ubuntu-latest**, which cannot compile the Cocoa universal Mach-O — the pom clang step is gated to the `mac` profile (inactive on Linux), so the Mac release assembly would fail (`target/RouteConverter` missing; the old shell launcher was deleted).

**Fix (mirrors the hosted-JRE pattern):**
- `build-mac-jre.yml` (macos-latest): compile the universal (`x86_64`+`arm64`) stub from `mac-launcher/JavaAppLauncher.m`, host `macos-launcher-stub` (+ `.sha256`) next to the stripped JREs. Added the source to the trigger paths.
- `_build-linux-mac.yml` (ubuntu): fetch + sha256-verify the stub, `install -m 0755` it to `RouteConverterMac/target/RouteConverter` and `TimeAlbumProMac/target/TimeAlbumPro` before `mvn package`. One byte-identical stub serves both apps (it resolves jar/icon/name from its own bundle at runtime).
- The pom `mac`-profile clang step stays as the **local-mac** build path (untouched).

**Bootstrap:** on merge, `build-mac-jre.yml` is in the changed paths so it reruns and publishes the stub; the prerelease must be re-run after that completes.

Verified locally: clang produces a universal binary; both workflow YAMLs parse. Signing/notarization of the stub via `_sign-mac.yml` runs only on the macOS release job (unchanged, inner-out codesign covers `Contents/MacOS/<App>`).

PR #217, opened 2026-07-23
fix(mac): run .app assembly on Linux runner (unblocks prerelease)

cpesch:
Third and final piece of the macOS launcher fix. The prerelease was still red after #216: `cp: cannot stat RouteConverterMac/target/RouteConverterMac-x64-app.zip`.

**Cause:** #208 put `maven-assembly-plugin` **inside** the `mac-native-launcher` (`os family=mac`) profile, together with the clang step. But the release `.app` zips are assembled in `_build-linux-mac.yml` on **ubuntu-latest**, where that profile is inactive — so the assembly `single` executions never ran (the module stopped at `shade`), no app-zip was produced, and staging failed. Affected both `RouteConverterMac` and `TimeAlbumProMac`.

**Fix:** move `maven-assembly-plugin` back to the main `<build>` (runs on any host); keep **only** the clang compile step behind the mac gate. On CI the Linux runner fetches the hosted universal stub into `target/<App>` (build-mac-jre.yml host + the #216 fetch step); on a local macOS build the profile compiles it into the same path.

**Verified locally** by simulating the Linux path (`-P '!mac-native-launcher'` + seeded stub/Runtimes): both `RouteConverterMac-x64-app.zip` and `-aarch64-app.zip` are produced, each with a universal (`x86_64`+`arm64`) `Contents/MacOS/RouteConverter`.

### DIFF
--- a/.github/workflows/_build-linux-mac.yml
+++ b/.github/workflows/_build-linux-mac.yml
@@ -100,7 +100,10 @@ jobs:
         # value the inline jobs used before Phase 5.2.
         # No `clean` — runner is ephemeral; `clean` would also wipe the
         # Runtime-<arch> dirs unpacked from the hosted Mac JREs above.
-        run: ./mvnw --batch-mode -Prouteconverter -Dmaven.build.number=${{ github.run_number }} -DcrashReportSecret=${{ secrets.CRASH_REPORT_HMAC_SECRET }} package
+        # -DmacApp activates the mac-app profile in RouteConverterMac/TimeAlbumProMac
+        # so the .app bundles are assembled here (stub fetched above). The plain
+        # `mvn verify` CI omits it, so it does not fail on the absent stub.
+        run: ./mvnw --batch-mode -Prouteconverter -DmacApp -Dmaven.build.number=${{ github.run_number }} -DcrashReportSecret=${{ secrets.CRASH_REPORT_HMAC_SECRET }} package
       - name: Verify runtime on stripped JRE (class-load gate)
         # Reproduces the minimized-JRE condition from scripts/jre-modules.txt and
         # fails the build on a ClassNotFoundException/NoClassDefFoundError that a
--- a/RouteConverterMac/pom.xml
+++ b/RouteConverterMac/pom.xml
@@ -74,22 +74,56 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Assemble the macos-{x64,aarch64}.xml .app bundles. This runs on
+                 the LINUX assembly runner (_build-linux-mac.yml), which is where
+                 the release .app zips are produced, so it must NOT be gated to a
+                 macOS host. The universal Contents/MacOS/RouteConverter stub it
+                 installs is fetched from static.routeconverter.com into
+                 target/RouteConverter by that workflow (built on macOS in
+                 build-mac-jre.yml); on a local macOS build the mac-native-launcher
+                 profile below compiles it into the same path instead. -->
+            <plugin>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>x64</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                        <configuration>
+                            <descriptors>
+                                <descriptor>${basedir}/src/main/assembly/macos-x64.xml</descriptor>
+                            </descriptors>
+                            <finalName>RouteConverterMac</finalName>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>aarch64</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                        <configuration>
+                            <descriptors>
+                                <descriptor>${basedir}/src/main/assembly/macos-aarch64.xml</descriptor>
+                            </descriptors>
+                            <finalName>RouteConverterMac</finalName>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
     <profiles>
         <profile>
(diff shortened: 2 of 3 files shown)


cpesch:

Mechanically sound refactor (moves `maven-assembly-plugin` out of the mac-only profile into the unconditional `<build>`, leaving only the clang compile step gated) that matches the stated intent of fixing the Linux release-assembly job. However, doing this makes the assembly executions run on every `package`-phase build, not just the release workflow that fetches the launcher stub — flagged as a likely regression risk for the ordinary CI test matrix.

### RouteConverterMac/pom.xml
- **major** [correctness] line 77 — `maven-assembly-plugin` now runs unconditionally in the package phase (previously gated behind the `mac-native-launcher` os-family profile), so any ordinary `mvn package`/`mvn verify` run — not just the release `_build-linux-mac.yml` job that fetches the hosted launcher stub — will now attempt to assemble the `.app` zip and likely fail hard if `target/RouteConverterMac` isn't present, instead of the previous silent skip.

### TimeAlbumProMac/pom.xml
- **major** [correctness] line 80 — Same issue as `RouteConverterMac/pom.xml`: the assembly plugin executions are no longer gated by the mac-only profile, so the ordinary CI test matrix build (if it also runs through the package phase for this module) will now require `target/TimeAlbumPro` to exist or fail, where it previously just skipped app-zip assembly.

_Please confirm (e.g. by pointing to the relevant workflow, or a green run of the standard `mvn verify` matrix job on this branch) that the non-release CI path either doesn't reach the package phase for these modules, or also has the stub available — otherwise this looks like it trades a red release job for a red standard-CI job._

---

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: Based on my review, this branch already contains a complete and correct fix for the described failure, and I'm making no edits.

## Investigation

**Root cause (per PR #217 description):** `#208` bound `maven-assembly-plugin`'s `single` executions (producing the `.app` zips) inside the `mac-native-launcher` profile (`os family=mac`). The release `.app` zips are actually assembled on `ubuntu-latest` in `_build-linux-mac.yml`, where that profile never activates, so the assembly step silently never ran and the later `cp RouteConverterMac-x64-app.zip` staging step failed with "cannot stat".

**Current state:** commit `49dca2770` (already at HEAD, working tree clean) moves `maven-assembly-plugin` out of the `mac-native-launcher` profile into each module's main `<build>` in both `RouteConverterMac/pom.xml` and `TimeAlbumProMac/pom.xml`, leaving only the `clang` compile step (`exec-maven-plugin`, bound to `prepare-package`) behind the mac-family gate.

I verified the mechanics that make this correct rather than just accepting the commit message:

- **Phase ordering (macOS local build):** the profile's `exec-maven-plugin` execution is bound to `prepare-package`, which runs before `package` regardless of plugin declaration order in the merged POM, so `target/RouteConverter` (or `target/TimeAlbumPro`) exists before the assembly `single` goals run.
- **Phase ordering (Linux CI build):** `maven-shade-plugin` (declared first) and `maven-assembly-plugin` (declared second) are both bound to `package`; same-phase executions run in declaration order, so the shaded jar (`target/RouteConverterMac.jar`) exists before assembly consumes it in the descriptor.
- **Path consistency:** `_build-linux-mac.yml`'s "Fetch hosted macOS launcher stub" step installs to `RouteConverterMac/target/RouteConverter` / `TimeAlbumProMac/target/TimeAlbumPro`, and unzips the hosted JRE into `target/Runtime-{x64,aarch64}` — both match exactly what `src/main/assembly/macos-{x64,aarch64}.xml` expects as `<source
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-13466` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-13466).

cpesch (commented):
## Verdict: approve · confidence 0.86

Moves `maven-assembly-plugin` out of the OS-family-gated `mac-native-launcher` profile into a new property-activated `mac-app` profile (`-DmacApp`), decoupling app-bundle assembly from the macOS-only clang compile step so it can run on the Linux release runner. The pom.xml diffs are internally consistent — the clang-compile plugin block and the (unchanged) assembly plugin block are cleanly split into two adjacent profiles — and `_build-linux-mac.yml` now passes `-DmacApp` to activate the new profile, while plain `mvn verify` CI still omits it. Author reports local verification of both arch app zips with universal binaries.

One behavioral note worth being aware of (already documented in the new comments): a local macOS build must now also pass `-DmacApp` explicitly to get an assembled `.app` — previously the OS-family gate alone was sufficient. Not a blocker, just a workflow change for local devs.

---

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: The task context contains an Inspector **approval** ("Verdict: approve · confidence 0.86"), not a change request. The one caveat it raises — that a local macOS build now needs `-DmacApp` explicitly — is already called out explicitly in the new `mac-app` profile's comment in `RouteConverterMac/pom.xml` (and mirrored in `TimeAlbumProMac/pom.xml`), and the feedback itself says "Not a blocker, just a workflow change for local devs."

There is no actionable change request here, so I'm making no edits.

No changes needed: Inspector approved PR 217 with no required follow-up

The Inspector feedback for PR 217 is an approval (confidence 0.86), not a change request — it explicitly says the one behavioral note (local macOS builds must now pass `-DmacApp`) is "not a blocker." I verified that note is already documented in the new `mac-app` profile comments in both `RouteConverterMac/pom.xml` and `TimeAlbumProMac/pom.xml` and in `_build-linux-mac.yml`. Working tree is unchanged.
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-13495` (http://127.0.0.1:3000/rc/rc-meta/src/branch/agent/patcher/dryrun-13495).

PR #221, opened 2026-07-23
chore(build): mac .app smoke-gate + runner-safety doc + drop stale javadoc param

cpesch:
Three post-mortem hardening items from the macOS Open-with / prerelease repair (issues #206, PRs #208/#215/#216/#217/#220):

1. **Smoke-gate the assembled `.app` bundles** (`_build-linux-mac.yml`): after assembly, assert each of the four app-zips contains `Contents/MacOS/<App>` + `bin/java` + `Java/*.jar`, and that the launcher stub is a universal `x86_64`+`arm64` Mach-O (checked with `file` — `lipo` is macOS-only, this runs on ubuntu). Turns a silent missing/mislocated stub into a loud build failure instead of a downstream `cp` error or a broken shipped app.
2. **Runner-safety doc** (`mac-launcher/README.md`): what the stub is + how it is built/hosted/fetched, plus a checklist for hand-rolled Maven/CI changes (profile activation × runner OS, which workflow runs it, `target/` phase) and the simulate-the-runner verification commands. The assembly-in-`os=mac`-profile trap was invisible three ways.
3. **Drop the stale `<additionalparam>-Xdoclint:none</additionalparam>`** (`pom.xml`, both javadoc configs): the parameter was removed in maven-javadoc-plugin 3.x (logged "unknown parameter") and the modern `<doclint>none</doclint>` was already present alongside it — so this just removes the redundant line. No behaviour change.

Verified locally: both workflow YAML + pom parse; the `file`-based universal check matches a real fat Mach-O; `<doclint>none</doclint>` retained (x2), `additionalparam` gone.

### DIFF
--- a/.github/workflows/_build-linux-mac.yml
+++ b/.github/workflows/_build-linux-mac.yml
@@ -114,6 +114,36 @@ jobs:
           scripts/verify-runtime.sh \
             RouteConverterLinux/target/RouteConverterLinux.jar \
             RouteConverterCmdLine/target/RouteConverterCmdLine.jar
+      - name: Smoke-check assembled .app bundles
+        # Fail loudly here (not at a downstream cp / at runtime) if a bundle is
+        # missing its universal launcher stub, JRE, or jar -- the failure modes
+        # behind the 2026-07 Open-with regression + the mislocated assembly.
+        # Runs on ubuntu, so use `file` (not the macOS-only `lipo`) to confirm
+        # the stub is a fat Mach-O with both arches.
+        run: |
+          set -euo pipefail
+          check() {  # $1 = app-zip, $2 = executable/app name
+            local zip="$1" exe="$2" app="$2.app" t
+            echo "== $zip =="
+            unzip -l "$zip" | grep -qE "${app}/Contents/MacOS/${exe}\$"  || { echo "::error::$zip: missing MacOS/${exe}"; exit 1; }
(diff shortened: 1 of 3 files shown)


cpesch (commented):

Maintainer PR adding a Mac .app bundle smoke-check to CI, a runner-safety README for the launcher stub, and removal of a redundant stale javadoc param. No blockers found; two nit-level robustness notes on the new smoke-check shell function.

### .github/workflows/_build-linux-mac.yml

- **nit** [correctness] line 130 — The `${app}` variable (e.g. `RouteConverter.app`) is interpolated unescaped into `grep -E` patterns, so the literal `.` in `.app` acts as a regex any-char wildcard rather than a literal dot.
```suggestion
local app_re="${app//./\\.}"
            unzip -l "$zip" | grep -qE "${app_re}/Contents/MacOS/${exe}\$"  || { echo "::error::$zip: missing MacOS/${exe}"; exit 1; }
            unzip -l "$zip" | grep -qE "${app_re}/Contents/bin/java\$"       || { echo "::error::$zip: missing bin/java"; exit 1; }
            unzip -l "$zip" | grep -qE "${app_re}/Contents/Java/.+\.jar\$"   || { echo "::error::$zip: missing Java/*.jar"; exit 1; }
```
- **nit** [style] line 138 — The `mktemp -d` temp dir used to extract the launcher stub is never removed; harmless on an ephemeral runner but leaves cleanup implicit.

---

Issue #222, opened 2026-07-23
Sort the map dropdown alphabetically (offline block, then online block)


cpesch:
## Feature: Sort the map dropdown alphabetically (offline block, then online block)

**Source:** Maintainer request (2026-07-23), from a screenshot of the "Karte:" map dropdown showing online maps in arbitrary catalog order (Outdoor Topo, KompassOSM, Bergfex OEK, Thunderforest Transport, Viamichelin, ArcGIS Topo, ArcGIS Satellite, Mapbox Satellite …). Goal: make the map dropdown predictable — each maps list sorted alphabetically by its displayed name.
This spec is fully resolved (grilled 2026-07-23); implement without further questions.

### Scope
Sort the two map lists that feed the "Karte:" combo (`getAvailableMapsModel()` in `MapsforgeMapManager`) **alphabetically by `description()`, case-insensitively**, and keep them sorted as items are added dynamically (offline maps discovered from directories, online maps built from the catalog, downloaded maps added later).

Two **independent** alphabetical blocks, in this order:
1. **All offline maps, A–Z** (`availableOfflineMapsModel`)
2. **All online maps, A–Z** (`availableOnlineMapsModel`)

This matches the existing `JoinedItemTableModel<>(availableOfflineMapsModel, filtered(availableOnlineMapsModel))` structure — it already renders the offline sub-model's rows first, then the online sub-model's rows. Sort each sub-model independently; do **not** merge into one combined list.

**Pure alphabetical — `OpenStreetMap` is NOT pinned.** It sorts under "O" like any other entry. No special casing for the default map in the ordering.

**Out of scope:**
- Themes / theme-styles / theme-style-categories models (already sorted via `Comparator.comparing(...description)` before insert — leave as is).
- Downloadable maps/themes models (`downloadableMapsModel`, `downloadableThemesModel`) — the maintainer already sorts those explicitly at populate time (`MapsforgeMapManager` ~L323/330); do not change.
- Any change to catalog/server-side ordering. This is a client-only display change.
- The Options-dialog tile-server table shares `availableOnlineMapsModel`, so it will also show the online maps sorted A–Z. That is acceptable/expected — no separate handling.

### Files
[...]

PR #225, opened 2026-07-23
Sort the map dropdown alphabetically (offline block, then online block)

cpesch:
All four files from the spec are present. Implementation complete.

## Summary

Implemented alphabetical sorting for the "Karte:" map dropdown (offline block, then online block) per issue #222.

**Changes:**

1. **`tileserver-maps/src/main/java/slash/navigation/maps/item/ItemTableModel.java`** — Added an optional `Comparator<? super T>` field with a new constructor `ItemTableModel(int columnCount, Comparator<? super T> comparator)`. The existing `ItemTableModel(int columnCount)` delegates with `comparator = null`, preserving append-only behavior for themes/styles/downloads. Extracted the insertion-index calculation to a pure static seam `sortedIndex(items, item, comparator)`; `addItem` uses it for a sorted insert (`items.add(index, item)` + `fireTableRowsInserted(index, index)`) when a comparator is set, otherwise falls back to the original append logic unchanged.

2. **`mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/models/TileMapTableModel.java`** — Passes `Comparator.comparing(Item::description, String.CASE_INSENSITIVE_ORDER)` to `super(...)`, so the online-maps model (`availableOnlineMapsModel`) self-sorts as maps are added by the catalog mediator.

3. **`mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/MapsforgeMapManager.java`** — Constructs `availableOfflineMapsModel` with the same case-insensitive `description()` comparator, so offline maps discovered from directories self-sort.

4. **`tileserver-maps/src/test/java/slash/navigation/maps/item/ItemTableModelSortTest.java`** (new) — JUnit tests covering the `sortedIndex` seam: out-of-order inserts end up sorted, insert into empty model, insert at start/middle/end, duplicate descriptions land stably after the existing equal item, and the no-comparator constructor preserves plain insertion order.

Both `availableOfflineMapsModel` and `availableOnlineMapsModel` are sorted independently (not merged), matching the existing `JoinedItemTableModel` structure that already renders offline rows first, then online rows. `OpenStreetMap` is not pinned — it sorts purely under "O". Themes/styles/downloadable-maps models are untouched (out of scope per the issue).

Couldn't run `mvn`/`javac` in this sandbox (blocked), so changes were verified by careful manual review against the existing code conventions rather than an actual build/test run.

Source issue: cpesch/RouteConverter#222

---
**Builder stats** · model `sonnet` · confidence `0` · 2m 10s across 1 round(s) · 4 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/13597)

### DIFF
--- a/mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/MapsforgeMapManager.java
+++ b/mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/MapsforgeMapManager.java
@@ -25,6 +25,7 @@ import slash.navigation.datasources.DataSourceManager;
 import slash.navigation.datasources.Downloadable;
 import slash.navigation.download.Download;
 import slash.navigation.gui.models.FilteringTableModel;
+import slash.navigation.maps.item.Item;
 import slash.navigation.maps.item.ItemModel;
 import slash.navigation.maps.item.ItemTableModel;
 import slash.navigation.maps.mapsforge.helpers.ActiveTileMapPredicate;
@@ -80,7 +81,8 @@ public class MapsforgeMapManager {
 
     private final DataSourceManager dataSourceManager;
     private final ItemTableModel<TileDownloadMap> availableOnlineMapsModel = new TileMapTableModel();
-    private final ItemTableModel<LocalMap> availableOfflineMapsModel = new ItemTableModel<>(1);
+    private final ItemTableModel<LocalMap> availableOfflineMapsModel = new ItemTableModel<>(1,
+            Comparator.comparing(Item::description, String.CASE_INSENSITIVE_ORDER));
     private final JoinedItemTableModel<LocalMap> availableMapsModel = new JoinedItemTableModel<>(availableOfflineMapsModel,
             new FilteringTableModel<>(availableOnlineMapsModel, new ActiveTileMapPredicate()));
     private final ItemTableModel<LocalTheme> availableThemesModel = new ItemTableModel<>(1);
--- a/mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/models/TileMapTableModel.java
+++ b/mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/models/TileMapTableModel.java
@@ -19,10 +19,12 @@
 */
 package slash.navigation.maps.mapsforge.models;
 
+import slash.navigation.maps.item.Item;
 import slash.navigation.maps.item.ItemTableModel;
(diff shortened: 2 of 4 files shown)


cpesch:

Sorted-insertion for the offline/online map dropdown is implemented cleanly (shared comparator seam in ItemTableModel, decent new unit test coverage), but one changed file calls `Comparator.comparing` without adding the import that its sibling file added for the same change — worth a local build check since the author could not run `mvn`/`javac` in this sandbox. One minor performance nit on the new insertion helper.

### mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/MapsforgeMapManager.java

- **major** [correctness] line 84 — This hunk calls `Comparator.comparing(...)` but the diff only adds an import for `Item`, not for `java.util.Comparator` — the sibling file (`TileMapTableModel.java`) needed that exact import for the identical change. Unless `java.util.Comparator` is already imported elsewhere in this file, this will fail to compile. The PR body states `javac`/`mvn` could not be run in the sandbox, so please confirm a local build (`mvn -pl mapsforge-maps compile`) succeeds before merge.

### tileserver-maps/src/main/java/slash/navigation/maps/item/ItemTableModel.java

- **minor** [style] line 95 — `sortedIndex` does a linear O(n) scan per insert, so bulk-loading a catalog (directory scan / online-catalog sync) is O(n^2). Probably fine at current map-list sizes, but if catalogs grow, `Collections.binarySearch` with an insertion-point fallback would avoid the quadratic behavior.

---

cpesch:
🤖 Patcher pushed a fix to `agent/builder/222-sort-the-map-dropdown-alphabetically-off`.

cpesch:

Sorted-insert implementation and tests are correct and well-targeted, but MapsforgeMapManager.java uses Comparator.comparing(...) without adding a java.util.Comparator import, unlike the other two touched files that explicitly add it — likely a compile failure the author's own PR description flags as unverified (mvn/javac blocked in sandbox).

### mapsforge-maps/src/main/java/slash/navigation/maps/mapsforge/MapsforgeMapManager.java
- **major** [correctness] line 85 — Comparator.comparing(...) is used here but the diff only adds `import slash.navigation.maps.item.Item;` — no `import java.util.Comparator;` is added, unlike the equivalent change in TileMapTableModel.java which explicitly adds that import. If this class doesn't already import java.util.Comparator elsewhere, this will fail to compile.
```suggestion
import java.util.Comparator;
```

---

bracki:
Haha @cpesch, still going strong here 💯

PR #232, opened 2026-07-28
ci(gitea): add verify.yml for factory PR verification (rc-meta#200)

cpesch:
Adds `.gitea/workflows/verify.yml` — inert on GitHub (only `.github/workflows` is read here), consumed by Gitea Actions on the private factory workspace `rc/RouteConverter` after the daily fork-tracking sync (which is ff-only, so the file must land here, not on the Gitea side).

Runs `./mvnw -B -ntp -DskipITs test` on the no-secrets `verify` runner for every PR (and on explicit factory dispatch after builder/patcher pushes). Design: rc-meta specs/00069-factory-computational-verification.md.

### DIFF
--- a/.gitea/workflows/verify.yml
+++ b/.gitea/workflows/verify.yml
@@ -1,0 +1,82 @@
+# PR verification for factory workspaces (Option A of rc-meta#200,
+# specs/00069-factory-computational-verification.md).
+#
+# CANONICAL COPY lives in rc-meta at factory/verify/verify.yml.template.
+# Deployed to <target repo>/.gitea/workflows/verify.yml (e.g.
+# rc/RouteConverter). Keep the two in sync — edit here, then re-deploy.
+#
+# Runs the repo's own test suite on the `verify` runner label: a separate
+# act_runner instance whose job container (verify-runner image) carries NO
+# secrets — see bootstrap-server/render-act-runner-verify-config.sh. The
+# code under test is agent-authored (untrusted); that is why:
+#   - checkout uses persist-credentials: false (no token left in .git/config)
+#   - the mvn step gets no secrets in env
+#   - the workflow-level GITEA_TOKEN is only referenced in the report step
+#
+# Triggers:
+#   - pull_request: human-pushed PRs (maintainer, manual branches)
+#   - workflow_dispatch: the factory (builder/patcher) dispatches explicitly
+#     after pushing with its PAT — Actions-token-adjacent pushes cannot be
+#     relied on to fire events (loop guard; memory gitea-actions-token-push).
+#     Input `pr` carries the PR number for the failure comment.
+name: verify
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: 'PR number to report the verdict on (optional)'
+        required: false
+        default: ''
+
+concurrency:
+  group: verify-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: verify
+    timeout-minutes: 45
+    steps:
+      - name: checkout (no persisted credentials)
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: run test suite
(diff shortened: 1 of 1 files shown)


cpesch:

Well-documented, minimal single-file addition implementing the no-secrets verify runner per specs/00069; one finding below on an injection anti-pattern that's explicitly called out in AGENTS.md.

### .gitea/workflows/verify.yml

- **major** [security] line 71 — `PR='${{ github.event.inputs.pr || github.event.pull_request.number }}'` inlines a GitHub Actions expression directly into a `run:` shell command; a stray quote/metacharacter in the optional `pr` workflow_dispatch string input breaks the quoting (or worse) before any guard runs — the exact pattern AGENTS.md's injection rule prohibits. The jq `--arg` body two lines later shows the safe pattern is already known in this file.
```suggestion
        env:
          GITEA_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          PR_INPUT: ${{ github.event.inputs.pr }}
          PR_EVENT: ${{ github.event.pull_request.number }}
        run: |
          set -uo pipefail
          PR="${PR_INPUT:-$PR_EVENT}"
          [[ -n "$PR" ]] || { echo "no PR number — skip comment"; exit 0; }
```

---

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: No safe fix identified: empty CI log and no reproducible failure signal

**Investigation summary**

- The branch `gitea-verify-ci` is exactly `master` plus one commit (`d326edc5b`) that adds `.gitea/workflows/verify.yml` — nothing else. `git diff master gitea-verify-ci --stat` confirms only that one new file.
- `.gitea/workflows/**` is not read by GitHub Actions (only `.github/workflows/**` is), so this PR's actual content is inert for every GitHub-side workflow — it can't have broken `build.yml`, `actionlint.yml` (path-filtered to `.github/workflows/**` only), or `workflow-config-guard.yml` (path-filtered to specific release/signing workflow files).
- `build.yml` (the only workflow that runs unconditionally on `push`) would therefore produce, on this branch, the exact same result it produces on `master` at the same commit — i.e. any failure here is not something introduced by this PR's diff.
- Checked the two dependency-bump commits already inherited from `master` (`jaxb2-maven-plugin` 3.3.0→4.1.0, and the `maven-minor-patch` group bump) as a possible pre-existing regression: the project already builds against `jakarta.xml.bind`/`jakarta.xml.bind-api` 4.0.x (`common/pom.xml`, `download-gui/pom.xml`, `download-tools/pom.xml`), so the plugin bump isn't an obvious namespace/API break, and I found no other config in the six `jaxb2-maven-plugin` consumers (`navigation-formats`, `download`, `datasource`, `geonames`, `route-catalog`, `tileserver-maps`) using parameters that changed between those versions.
- The `agentic-task.json` context's `log_tail` is empty, so there is no actual error text to diagnose against.
- The sandbox has no `java`/`mvn` on `PATH` and no network access, so I cannot execute `./mvnw -DskipITs test` (or `verify`) here to reproduce/observe a real failure, nor fetch the linked Actions run (`.../runs/30390979857`) to read its log.

**Conclusion**

With no failure log content, a PR diff that is provably inert to every GitHub-triggered workflow, and no 
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-14874` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-14874).

cpesch:
INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   Gpx11ExtensionsTest.testUpdateHdopDomElementInPlace:912
[ERROR]   Gpx11ExtensionsTest.testUpdateSpeedUpdatesAllExistingRepresentationsConsistently:977
[INFO] 
[ERROR] Tests run: 682, Failures: 2, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------

cpesch:
🤖 Patcher pushed a fix to `gitea-verify-ci`.

PR #244, opened 2026-08-01
Potential fix for code scanning alert no. 1: Cache Poisoning via low-privileged code injection

cpesch:
Potential fix for [https://github.com/cpesch/RouteConverter/security/code-scanning/1](https://github.com/cpesch/RouteConverter/security/code-scanning/1)

General fix: do not inline untrusted GitHub expression values inside shell scripts. Instead, pass them through `env:` and read them as normal shell variables (`$VAR`) with proper quoting.

Best fix here (minimal, no functional change): in `.github/workflows/verify-cut.yml`, update the `Resolve tag from triggering release run` step to:
- add environment variables for `HEAD_BRANCH` and `RUN_CONCLUSION` sourced from GitHub context
- assign `TAG` and `CONCLUSION` from those env vars inside the script
- keep all existing outputs/echo behavior unchanged

This removes direct expression injection into `run:` while preserving current logic.

### DIFF
--- a/.github/workflows/verify-cut.yml
+++ b/.github/workflows/verify-cut.yml
@@ -16,9 +16,11 @@ jobs:
         id: ctx
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+          RUN_CONCLUSION: ${{ github.event.workflow_run.conclusion }}
         run: |
-          TAG='${{ github.event.workflow_run.head_branch }}'
-          CONCLUSION='${{ github.event.workflow_run.conclusion }}'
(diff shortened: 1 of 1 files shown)


cpesch (commented):
## Verdict: approve · confidence 0.95

No blockers found. Security fix correctly refactors the workflow step to pass GitHub expression values through environment variables rather than inlining them into the shell script. This eliminates the injection surface that the code scanning alert flagged — a malicious `head_branch` or `conclusion` value could previously break out of single quotes and execute arbitrary commands.

**The fix:**
- Moves `${{ github.event.workflow_run.head_branch }}` and `${{ github.event.workflow_run.conclusion }}` to `env:` keys
- Reads them as shell variables with proper double-quote expansion
- Preserves all outputs and echo behavior

No convention violations detected. Ready to merge.

---

PR #245, opened 2026-08-01
Potential fix for code scanning alert no. 1: Cache Poisoning via low-privileged code injection

cpesch:
Potential fix for [https://github.com/cpesch/RouteConverter/security/code-scanning/1](https://github.com/cpesch/RouteConverter/security/code-scanning/1)

General fix: do not inline untrusted GitHub expression values inside shell scripts. Instead, pass them through `env:` and read them as normal shell variables (`$VAR`) with proper quoting.

Best fix here (minimal, no functional change): in `.github/workflows/verify-cut.yml`, update the `Resolve tag from triggering release run` step to:
- add environment variables for `HEAD_BRANCH` and `RUN_CONCLUSION` sourced from GitHub context
- assign `TAG` and `CONCLUSION` from those env vars inside the script
- keep all existing outputs/echo behavior unchanged

This removes direct expression injection into `run:` while preserving current logic.

### DIFF
--- a/.github/workflows/verify-cut.yml
+++ b/.github/workflows/verify-cut.yml
@@ -16,9 +16,11 @@ jobs:
         id: ctx
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+          RUN_CONCLUSION: ${{ github.event.workflow_run.conclusion }}
         run: |
-          TAG='${{ github.event.workflow_run.head_branch }}'
-          CONCLUSION='${{ github.event.workflow_run.conclusion }}'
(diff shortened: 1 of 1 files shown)


cpesch (commented):
## Verdict: approve · confidence 0.98

This is a correct security hardening fix. Moving `${{ github.event.workflow_run.head_branch }}` and `${{ github.event.workflow_run.conclusion }}` from inline script interpolation into environment variables, then referencing them with proper quoting, eliminates shell injection if either property contains special characters (particularly single quotes, which would break the original script's quoting).

The env-var pattern is the standard GitHub Actions mitigation for this class of injection and is correctly applied here.

---

cpesch:
Duplicate of #244 (identical fix for the same alert). Closing in favor of #244, which merged.

PR #253, opened 2026-08-01
fix(navigation-formats): replace ColumbusV1000Device's shared-OS-Preferences test race with in-memory fake

### LINKED ISSUE #252: [follow-up #251] Fix Columbus timezone test race root cause (in-memory Preferences fake, revert forkCount=1)
## Refactor: replace ColumbusV1000Device's shared-OS-Preferences test race with an in-memory fake, then revert the forkCount=1 pin

**Source:** follow-up to merged PR [cpesch/RouteConverter#251](https://github.com/cpesch/RouteConverter/pull/251) (`b9ebda536`). That PR's own inspector review (confidence 0.85) flagged its fix as a workaround: pinning `forkCount=1` on the whole `navigation-formats` module serializes its *entire* test suite in one JVM just to stop a handful of Columbus tests racing on shared OS state, forfeiting parallelism for everything else in the module. Goal here: fix the actual root cause using the pattern this codebase already proved out for the identical problem (PR #140 / `IntegerModelTest`), then revert the forkCount pin.

Root cause (unchanged from #251's diagnosis): `ColumbusV1000Device`'s `useLocalTimeZone`/`timeZone` flags are backed by `Preferences.userNodeForPackage(ColumbusV1000Device.class)` — a real, process-external, OS-shared store (Windows Registry / `~/.java/.userPrefs` / plist), not JVM-local state. Under the root pom's default `forkCount=1.5C`, two Columbus test classes can land in different forked JVMs at the same time and race on that one shared external node.

This spec is fully resolved (grilled 2026-08-01); implement without further questions.

### Scope

In scope:
- Give `ColumbusV1000Device` a swappable backing store (a test seam), instead of forcing single-JVM execution.
- Relocate the existing `InMemoryPreferences` fake (`common-gui/src/test/java/slash/navigation/gui/models/InMemoryPreferences.java`, added in #140) into the `common` module's test-jar, since `common-gui`, `route-converter-gui`, **and** `navigation-formats` all already depend on `common`'s test-jar — no new pom dependency needed anywhere. (`navigation-formats` sits below `common-gui` in the module graph; depending on `common-gui`'s test-jar from `navigation-formats` would be a backwards/upward dependency, which is why the fake moves down into `common` rather than `navigation-formats` reaching up into `common-gui`.)
- Wire the four Columbus test classes that actually touch the shared flags onto the fake.
- Revert the `forkCount=1` pin added by #251, restoring `navigation-formats` to the root pom's default `forkCount=1.5C`.

Out of scope (explicitly, don't touch):
- `ColumbusGpsType1FormatTest` — checked, it does **not** call `ColumbusV1000Device` or `ConvertBase.ignoreLocalTimeZone()` at all. It was in the original bug report's suspect list but isn't actually part of the race. No change.
- The five `*IT.java` classes that also call `ConvertBase.ignoreLocalTimeZone()` (`ExcelReadWriteRoundtripIT`, `AccurracyConvertIT`, `ConvertIT`, `ColumbusFusionReadWriteRoundtripIT`, `ColumbusGpsType2ReadWriteRoundtripIT`): they run under the Failsafe integration-test phase, not the Surefire unit-test phase where the race was observed (PR #243's CI symptom was on `*Test` classes only). Leave them reading the real OS Preferences store — no known race there, and swapping them in is scope creep beyond the reported bug.
- No change to `ConvertBase.ignoreLocalTimeZone()` itself — it just toggles `getUseLocalTimeZone()`/`setUseLocalTimeZone()` around a runnable and is agnostic to which store backs those calls.

### Files

**Main code**
- `navigation-formats/src/main/java/slash/navigation/columbus/ColumbusV1000Device.java`

**Fake relocation**
- DELETE `common-gui/src/test/java/slash/navigation/gui/models/InMemoryPreferences.java`
- NEW `common/src/test/java/slash/common/prefs/InMemoryPreferences.java` (identical content, package changed to `slash.common.prefs`)
- `common-gui/src/test/java/slash/navigation/gui/models/IntegerModelTest.java` (import only)
- `route-converter-gui/src/test/java/slash/navigation/converter/gui/models/RecentFormatsModelTest.java` (import only)
- `route-converter-gui/src/test/java/slash/navigation/converter/gui/models/RecentUrlsModelTest.java` (import only)
- `route-converter-gui/src/test/java/slash/navigation/converter/gui/models/RoutesTableSortPreferencesTest.java` (import only)
- `route-converter-gui/src/test/java/slash/navigation/converter/gui/helpers/AutomaticGeocodingServiceTest.java` (import only)
- `route-converter-gui/src/test/java/slash/navigation/converter/gui/helpers/GeocodingServiceFacadeTest.java` (import only)

**Columbus test wiring**
- `navigation-formats/src/test/java/slash/navigation/columbus/ColumbusV1000DeviceTest.java`
- `navigation-formats/src/test/java/slash/navigation/columbus/ColumbusGpsBinaryFormatTest.java`
- `navigation-formats/src/test/java/slash/navigation/columbus/ColumbusFusionFormatTest.java`
- `navigation-formats/src/test/java/slash/navigation/columbus/ColumbusGpsType2FormatTest.java`

**Build config**
- `navigation-formats/pom.xml`

1. **`ColumbusV1000Device` seam.** `ColumbusV1000Device` is a static utility with no instances, used from production code too (`BaseRouteConverter.java:1248`, `OptionsDialog.java:609-611`), so it can't take a constructor-injected `Preferences` the way `IntegerModel` does. Instead:
   - Change `private static final Preferences preferences = ...` to `private static Preferences preferences = ...` (drop `final`).
   - Add:
     ```java
     /**
      * Test seam — swaps the backing store so tests don't race on the shared OS-level
      * Preferences node under concurrent multi-fork test execution. Do not call from
      * production code.
      */
     public static void setPreferences(Preferences preferences) {
         ColumbusV1000Device.preferences = preferences;
     }
     ```
   - Public (not package-private): `ConvertBase.ignoreLocalTimeZone()` lives in `slash.navigation.base`, a different package from `ColumbusV1000Device` (`slash.navigation.columbus`), and calls `getUseLocalTimeZone()`/`setUseLocalTimeZone()` — but the swap itself happens in each test class's own `@Before`, in the `slash.navigation.columbus` package, so `ignoreLocalTimeZone()` doesn't need to call `setPreferences` itself, only the getters/setters it already uses.

2. **Fake relocation.** Move `InMemoryPreferences` verbatim (same `AbstractPreferences`/`ConcurrentHashMap` implementation) from `common-gui/src/test/java/slash/navigation/gui/models/` to `common/src/test/java/slash/common/prefs/`, changing only the `package` declaration. Update the `import slash.navigation.gui.models.InMemoryPreferences;` (or same-package unqualified use) in all six consumer test files listed above to `import slash.common.prefs.InMemoryPreferences;`. No pom.xml changes needed — `common-gui` and `route-converter-gui` already depend on `common`'s test-jar (confirmed: `common-gui/pom.xml`, `route-converter-gui/pom.xml`), and `navigation-formats/pom.xml` already depends on `common`'s test-jar too.

3. **Per-test isolation.** Each of the four Columbus test classes gets a fresh `InMemoryPreferences()` installed in `@Before`, replacing any save/restore boilerplate:
   - `ColumbusV1000DeviceTest`: replace the current `@Before`/`@After` (which capture `useLocalTimeZone`/`timeZone` via the getters and restore them in `@After`) with a single `@Before` that does `ColumbusV1000Device.setPreferences(new InMemoryPreferences());`. Delete the `tearDown()` method and the two capture fields (`useLocalTimeZone`, `timeZone`) — no longer needed since the fake store is thrown away after the test JVM/class anyway and the getters already default cleanly (`useLocalTimeZone` defaults `true`, `timeZone` defaults `TimeZone.getDefault().getID()`).
   - `ColumbusGpsBinaryFormatTest`: same pattern — its `@Before` currently does `useLocalTimeZone = ColumbusV1000Device.getUseLocalTimeZone(); ColumbusV1000Device.setUseLocalTimeZone(false);` and `@After` restores it. Replace with `@Before` that installs `new InMemoryPreferences()` first, then still calls `setUseLocalTimeZone(false)` (that part of the behavior — forcing UTC-relative parsing for the test — is unrelated to the race and must stay). Delete `@After`/`tearDown()` and the `useLocalTimeZone` field.
   - `ColumbusFusionFormatTest`: has no existing `@Before` — add one (new `import org.junit.Before;`) that installs `ColumbusV1000Device.setPreferences(new InMemoryPreferences());`. No `@After` needed.
   - `ColumbusGpsType2FormatTest`: same as `ColumbusFusionFormatTest` — add a `@Before` installing a fresh fake store; no `@After`.
   - `ColumbusGpsType1FormatTest`: **no change** (see Out of scope).

4. **pom.xml revert.** Revert both changes PR #251 made to `navigation-formats/pom.xml`, restoring it to its pre-#251 state exactly:
   - In the `local-with-samples` profile's `maven-surefire-plugin` `configuration` block: change `<forkCount>1</forkCount>` back to `<forkCount>1.5C</forkCount>`, and remove the `<!-- forkCount=1: see the forkCount note ... -->` comment above it. Leave `skipTests`, `combine.self="override"`, and the `<includes>` block untouched.
   - In `build/plugins`: delete the entire added `<plugin>` block for `maven-surefire-plugin` (the one with the long Columbus-timezone-race comment and a bare `<forkCount>1</forkCount>` configuration) — this whole block was net-new in #251 and doesn't exist upstream of it.

### Testable seam

No new pure function — this is a test-infrastructure fix. The "test" of correctness is the existing Columbus test suite passing under the restored `forkCount=1.5C` (i.e. running with real fork parallelism) reliably, repeatedly, without the intermittent 1-hour-offset failures PR #243 observed. No GUI involved.

### Acceptance criteria

- `mvn -pl navigation-formats,common,common-gui,route-converter-gui -am test` passes.
- `navigation-formats/pom.xml` no longer contains any `forkCount` override (diff against pre-#251 `navigation-formats/pom.xml` at commit `e16f8a2de` is empty for that file, i.e. `git diff e16f8a2de -- navigation-formats/pom.xml` shows nothing).
- Run the full `navigation-formats` Surefire suite 10x in a row locally (`for i in $(seq 10); do mvn -pl navigation-formats -am test -Dtest='Columbus*Test' || break; done`) with no intermittent timezone-offset failures — manual maintainer check, since the race was only ever reproducible on CI's "Java 21 on Windows" job across reruns.
- `common-gui/src/test/java/slash/navigation/gui/models/InMemoryPreferences.java` no longer exists; `common/src/test/java/slash/common/prefs/InMemoryPreferences.java` exists with identical logic.
- `grep -rn "ColumbusV1000Device.setPreferences" navigation-formats/src/test` shows exactly 4 call sites, one each in `ColumbusV1000DeviceTest`, `ColumbusGpsBinaryFormatTest`, `ColumbusFusionFormatTest`, `ColumbusGpsType2FormatTest`.

cpesch:
## Summary
- `ColumbusV1000Device`'s `useLocalTimeZone`/`timeZone` flags were backed by a real, process-external `Preferences.userNodeForPackage()` node; two Columbus test classes in different forked JVMs could race on that shared store (PR #251 papered over this with `forkCount=1` on the whole `navigation-formats` module).
- Adds a test-only `setPreferences()` seam on `ColumbusV1000Device`, moves the existing `InMemoryPreferences` fake (from #140) into `common`'s test-jar (avoids `navigation-formats` reaching up into `common-gui`), and wires the four affected Columbus test classes onto a fresh in-memory store per test.
- Reverts #251's `forkCount=1` pin now that the root cause is fixed — `navigation-formats` regains the root pom's default `forkCount=1.5C`.

Closes #252

## Test plan
- [x] `mvn -pl common,common-gui -am install -DskipTests` (fresh test-jar with relocated fake)
- [x] `mvn -pl navigation-formats test -Dtest='ColumbusV1000DeviceTest,ColumbusGpsBinaryFormatTest,ColumbusFusionFormatTest,ColumbusGpsType2FormatTest,ColumbusGpsType1FormatTest'` — all green
- [x] Full reactor build (`common,common-gui,navigation-formats,route-converter-gui`) — only pre-existing, unrelated failures remain (`ReadIT`/`ColumbusFusionReadWriteRoundtripIT` missing rc-samples fixtures, `UrlFormatTest` network flake); none touch the changed code
- [x] `git diff <pre-#251 commit> -- navigation-formats/pom.xml` is empty — forkCount revert is exact

### DIFF
--- a/common-gui/src/test/java/slash/navigation/gui/models/IntegerModelTest.java
+++ b/common-gui/src/test/java/slash/navigation/gui/models/IntegerModelTest.java
@@ -21,6 +21,7 @@
 package slash.navigation.gui.models;
 
 import org.junit.Test;
+import slash.common.prefs.InMemoryPreferences;
 
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
--- a/common/src/test/java/slash/common/prefs/InMemoryPreferences.java
+++ b/common/src/test/java/slash/common/prefs/InMemoryPreferences.java
@@ -18,7 +18,7 @@
     Copyright (C) 2007 Christian Pesch. All Rights Reserved.
 */
 
-package slash.navigation.gui.models;
+package slash.common.prefs;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
--- a/navigation-formats/pom.xml
+++ b/navigation-formats/pom.xml
@@ -187,9 +187,7 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration combine.self="override">
                             <skipTests>${skip.unit.tests}</skipTests>
-                            <!-- forkCount=1: see the forkCount note on this module's default
-                                 maven-surefire-plugin configuration above. -->
-                            <forkCount>1</forkCount>
+                            <forkCount>1.5C</forkCount>
                             <includes>
                                 <include>**/*Test.java</include>
                                 <include>**/*IT.java</include>
@@ -203,24 +201,6 @@
 
     <build>
         <plugins>
-            <plugin>
-                <!-- Several Columbus tests (ColumbusGpsBinaryFormatTest, ColumbusFusionFormatTest, ...)
-                     toggle ColumbusV1000Device's useLocalTimeZone/timeZone flags, which are backed by
-                     java.util.prefs.Preferences: a process-external, OS-shared store (Windows Registry,
-                     ~/.java/.userPrefs, ...), not JVM-local state. The root pom's forkCount=1.5C runs
-                     several test-class forks concurrently, so two of these tests can land in different
-                     forked JVMs at the same time and race on that shared store - one test's setUp/finally
-                     flips useLocalTimeZone/timeZone while another test's parseTime() is mid-assertion,
-                     producing an exact device-timezone-offset time shift (seen as e.g. Europe/Berlin's
-                     1-hour winter offset). forkCount=1 makes this module's tests run sequentially in one
-                     JVM, removing the race. See the "Java 21 on Windows" intermittent CI failures on
-                     PR #243 for the observed symptom. -->
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <forkCount>1</forkCount>
-                </configuration>
-            </plugin>
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
                 <artifactId>jaxb2-maven-plugin</artifactId>
--- a/navigation-formats/src/main/java/slash/navigation/columbus/ColumbusV1000Device.java
+++ b/navigation-formats/src/main/java/slash/navigation/columbus/ColumbusV1000Device.java
@@ -29,9 +29,18 @@ import java.util.prefs.Preferences;
  * @author Christian Pesch
  */
 public class ColumbusV1000Device {
-    private static final Preferences preferences = Preferences.userNodeForPackage(ColumbusV1000Device.class);
+    private static Preferences preferences = Preferences.userNodeForPackage(ColumbusV1000Device.class);
     private static final String TIMEZONE_PREFERENCE = "timeZone";
     private static final String USE_LOCAL_TIMEZONE_PREFERENCE = "useLocalTimeZone";
+
+    /**
+     * Test seam — swaps the backing store so tests don't race on the shared OS-level
+     * Preferences node under concurrent multi-fork test execution. Do not call from
+     * production code.
+     */
+    public static void setPreferences(Preferences preferences) {
+        ColumbusV1000Device.preferences = preferences;
+    }
 
     public static boolean getUseLocalTimeZone() {
         return preferences.getBoolean(USE_LOCAL_TIMEZONE_PREFERENCE, true);
--- a/navigation-formats/src/test/java/slash/navigation/columbus/ColumbusFusionFormatTest.java
+++ b/navigation-formats/src/test/java/slash/navigation/columbus/ColumbusFusionFormatTest.java
@@ -19,7 +19,9 @@
 */
 package slash.navigation.columbus;
 
+import org.junit.Before;
 import org.junit.Test;
+import slash.common.prefs.InMemoryPreferences;
 import slash.navigation.base.RouteCharacteristics;
 import slash.navigation.base.WaypointType;
 import slash.navigation.base.Wgs84Position;
@@ -43,6 +45,11 @@ import static slash.navigation.columbus.ColumbusV1000Device.setUseLocalTimeZone;
 
 public class ColumbusFusionFormatTest {
     private final ColumbusFusionFormat format = new ColumbusFusionFormat();
+
+    @Before
+    public void setUp() {
+        ColumbusV1000Device.setPreferences(new InMemoryPreferences());
+    }
 
     private static BufferedReader reader(String... lines) {
         StringBuilder builder = new StringBuilder();
--- a/navigation-formats/src/test/java/slash/navigation/columbus/ColumbusGpsBinaryFormatTest.java
+++ b/navigation-formats/src/test/java/slash/navigation/columbus/ColumbusGpsBinaryFormatTest.java
@@ -1,8 +1,8 @@
 package slash.navigation.columbus;
 
-import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
+import slash.common.prefs.InMemoryPreferences;
 
 import java.nio.ByteBuffer;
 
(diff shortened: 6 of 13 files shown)


common-gui/src/test/java/slash/navigation/gui/models/IntegerModelTest.java
```diff
@@ -21,6 +21,7 @@
 package slash.navigation.gui.models;
 
 import org.junit.Test;
+import slash.common.prefs.InMemoryPreferences;
```
cpesch, line 24: is that import needed?

cpesch:

Well-scoped test-isolation fix. The `setPreferences()` seam swaps `ColumbusV1000Device`'s backing store to a JVM-local `InMemoryPreferences` per test, which removes the cross-fork race on the OS-shared `java.util.prefs` node at its root and correctly justifies reverting #251's module-wide `forkCount=1` pin back to the root default. Save/restore `tearDown` blocks are dropped consistently with the per-test in-memory store, and the fake's relocation into `common`'s test-jar avoids `navigation-formats` reaching up into `common-gui`. The stated test plan exercises the four Columbus classes and the full reactor.

No blockers found. One non-blocking note (not visible in the diff, so not flagged as a finding): the new `slash.common.prefs.InMemoryPreferences` import in `navigation-formats` tests requires a `common` test-jar dependency in that module — the green `mvn -pl navigation-formats test` run implies it is already present.

---

PR #247, opened 2026-08-01
Potential fix for code scanning alert no. 6: Workflow does not contain permissions

cpesch:
Potential fix for [https://github.com/cpesch/RouteConverter/security/code-scanning/6](https://github.com/cpesch/RouteConverter/security/code-scanning/6)

Add an explicit top-level `permissions` block to `.github/workflows/build-windows-jre.yml` so all jobs inherit least-privilege defaults.

Best single fix (without changing functionality):
- Insert at workflow root (after `concurrency` and before `jobs`):
  - `permissions:`
  - `contents: read`

Why this is best:
- `actions/checkout` requires `contents: read`.
- Other actions used here (`setup-java`, artifact upload/download) do not require repository write permissions.
- Deployment uses SSH, not GitHub API writes.
- A root-level block is concise and applies to both `build` and `publish` consistently.

No new imports, methods, or dependencies are needed.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

PR #242, opened 2026-08-01
Potential fix for code scanning alert no. 4: Workflow does not contain permissions

cpesch:
Potential fix for [https://github.com/cpesch/RouteConverter/security/code-scanning/4](https://github.com/cpesch/RouteConverter/security/code-scanning/4)

Add an explicit top-level `permissions` block to `.github/workflows/build-windows-jre.yml` so all jobs inherit minimal `GITHUB_TOKEN` access.

Best fix (without changing behavior):
- Insert, near the top of the workflow (after `name:` and before `on:`),:
  - `permissions:`
  - `contents: read`

Why this is best:
- `actions/checkout` needs `contents: read`.
- The workflow does not create releases, comment on PRs, or otherwise require write scopes.
- Applying at workflow root covers both `build` and `publish` jobs consistently and minimally.

No imports/dependencies/methods are needed (YAML-only change).

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

PR #254, opened 2026-08-01
fix(ci): dedupe permissions key in build-windows-jre workflow

cpesch:
## Summary
- Two independent code-scanning autofix PRs (#242, #247) each added a top-level `permissions:` block to `build-windows-jre.yml`, producing a duplicate mapping key.
- Duplicate top-level keys are invalid YAML — GitHub rejects the workflow file outright, so every run since fails immediately with zero jobs (e.g. https://github.com/cpesch/RouteConverter/actions/runs/30707599676).
- Keep a single `permissions: contents: read` block.

## Test plan
- [ ] Merge and confirm the next `build-windows-jre` run (push to `pom.xml`/`scripts/jre-modules.txt`/this file, or manual `workflow_dispatch`) actually creates jobs and completes.

### DIFF
--- a/.github/workflows/build-windows-jre.yml
+++ b/.github/workflows/build-windows-jre.yml
@@ -34,9 +34,6 @@ concurrency:
   group: build-windows-jre
   cancel-in-progress: true
 
-permissions:
-  contents: read
-
 jobs:
   build:
     runs-on: windows-latest


cpesch (commented):
## Verdict: approve · confidence 1.0

Two autofix PRs (#242, #247) each independently added identical `permissions:` blocks, creating a YAML syntax error that breaks the entire workflow. This fix removes the duplicate, restoring validity. Change is minimal and correct.

---

PR #261, opened 2026-08-08
Skip crash-report upload when secret is the committed placeholder

### LINKED ISSUE #260: [maintainer] Skip crash-report upload when secret is the committed placeholder (spool never drains in dev builds)
## Bug: CrashReportSender posts placeholder-signed reports the server always rejects — spool never drains in dev builds

**Source:** maintainer session 2026-08-08, follow-up to #259's diagnosis. Observed: every dev/classpath launch logs "Sending 10 spooled crash report(s) from a previous session" and POSTs all spooled reports to `https://api.routeconverter.com/v1/crash-report/`; every POST fails signature verification because the effective secret is the committed placeholder (`WARNUNG: Crash report secret is the committed placeholder`). The spool (`CrashReportSpool`, cap 10) therefore never drains and the same 10 futile uploads repeat on every start.
This spec is fully resolved (grilled 2026-08-08); implement without further questions.

### Context (verified)

- Official builds inject the real key: `.github/workflows/_build-windows.yml` and `_build-linux-mac.yml` pass `-DcrashReportSecret=${{ secrets.CRASH_REPORT_HMAC_SECRET }}`, which Maven resource filtering writes into `crash-report.properties`. So the futile-upload behaviour affects **only dev/classpath runs** (no filtered resource, no `-Drouteconverter.crashReportSecret`), where `CrashReportSender.secret()` falls back to `DEFAULT_SECRET` (`"CHANGEME-crash-report-hmac-secret"`).
- `CrashReportSender.send(...)` currently warns once (`warnIfDefaultSecret()`) and then signs+POSTs anyway; the server rejects with non-2xx, `send` returns false, and `CrashReporter.offerSpooledReports()` keeps the file for the next launch.

### Scope

`feedback` module only: make `send` a no-op network-wise under the placeholder secret. Out of scope: any change to spooling behaviour (fresh crashes must keep being spooled — the on-disk spool is the primary local diagnostic, see #259), the spool cap, the server side, the warning text, and the build-time injection mechanism.

### Files

- `feedback/src/main/java/slash/navigation/feedback/domain/CrashReportSender.java`
- `feedback/src/test/java/slash/navigation/feedback/domain/CrashReportSenderTest.java`

1. In `CrashReportSender.send(String apiUrl, String json)`: after the existing oversized-payload check and the existing `warnIfDefaultSecret()` call, if `DEFAULT_SECRET.equals(secret())`, log at `FINE` ("Not sending crash report: placeholder secret, keeping it spooled") and return `false` **without** calling `sign(...)` or `post(...)`.
2. Returning `false` is deliberate: `CrashReporter` treats false as "keep the file spooled", which preserves the reports for local diagnosis and lets a later properly-configured run deliver them.
3. The one-time-per-run warning semantics of `warnIfDefaultSecret()` stay untouched (it already uses an `AtomicBoolean`).
4. No behaviour change when a real secret is configured (system property or filtered resource): sign + POST exactly as today.
5. No RELEASE_NOTES entry — dev-only behaviour, invisible in official builds.

### Testable seam (exists — extend, don't restructure)

`CrashReportSenderTest` (plain JUnit, `post` is overridable for transport substitution). Add:
- `testSendWithPlaceholderSecretDoesNotPost`: with no `routeconverter.crashReportSecret` system property set (and no bundled resource in the test classpath, i.e. `secret()` returns `DEFAULT_SECRET`), a sender whose overridden `post` records invocations must return `false` from `send(...)` and record **zero** `post` calls.
- `testSendWithConfiguredSecretStillPosts`: set the system property to a non-placeholder value (restore it in a finally/`@After`), stub `post` to return true → `send` returns true and `post` was called once with a non-empty signature header value.

All existing tests in the class must pass unmodified.

### Acceptance criteria

- `mvn -pl feedback test` green, including the two new tests (machine-verifiable).
- Manual maintainer check: dev classpath launch with spooled reports present shows the single placeholder warning, no `crash-report/` POSTs in the log, and the spool files still on disk afterwards.

cpesch:
Closes #260.

## Summary

- `CrashReportSender.send(...)` now checks, right after the existing one-time `warnIfDefaultSecret()` call, whether the effective secret is still `DEFAULT_SECRET`. If so, it logs at `FINE` ("Not sending crash report: placeholder secret, keeping it spooled") and returns `false` without calling `sign(...)` or `post(...)`.
- Added two tests to `CrashReportSenderTest`: `testSendWithPlaceholderSecretDoesNotPost` (no secret configured, `post` must not be invoked) and `testSendWithConfiguredSecretStillPosts` (secret configured, `post` is invoked with a non-empty signature).

## Why

Official builds inject the real HMAC key via Maven resource filtering, so this only affects dev/classpath launches. There, `secret()` falls back to the committed placeholder, `send` signed and POSTed anyway, the server always rejected the request, and `CrashReporter.offerSpooledReports()` kept re-attempting the same up-to-10 spooled reports on every launch. Returning `false` under the placeholder is exactly what `CrashReporter` already treats as "keep the file spooled" — so a fresh crash is still spooled, and a later run with a properly configured secret still delivers the backlog.

## Risk

Low. The change is additive and only short-circuits the network call when the secret is unmistakably the placeholder string; behaviour with a real secret (system property or filtered resource) is unchanged, and all existing tests in the class pass unmodified.

🤖 Builder.

---
**Builder stats** · model `sonnet` · confidence `0.95` · 1m 22s across 1 round(s) · 2 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/18653)

Closes #260

### DIFF
--- a/feedback/src/main/java/slash/navigation/feedback/domain/CrashReportSender.java
+++ b/feedback/src/main/java/slash/navigation/feedback/domain/CrashReportSender.java
@@ -89,6 +89,10 @@ public class CrashReportSender {
                 return false;
             }
             warnIfDefaultSecret();
+            if (DEFAULT_SECRET.equals(secret())) {
+                log.log(FINE, "Not sending crash report: placeholder secret, keeping it spooled");
+                return false;
+            }
             String signature = sign(secret(), body);
             return post(apiUrl + CRASH_REPORT_URI, json, signature);
         } catch (Exception e) {
--- a/feedback/src/test/java/slash/navigation/feedback/domain/CrashReportSenderTest.java
+++ b/feedback/src/test/java/slash/navigation/feedback/domain/CrashReportSenderTest.java
@@ -117,4 +117,39 @@ public class CrashReportSenderTest {
             System.clearProperty(SECRET_PROPERTY);
         }
     }
+
+    @Test
+    public void testSendWithPlaceholderSecretDoesNotPost() {
+        System.clearProperty(SECRET_PROPERTY);
+        CrashReportSender.resetDefaultSecretWarningForTesting();
+
+        final boolean[] posted = {false};
+        CrashReportSender sender = new CrashReportSender() {
+            boolean post(String url, String json, String signature) {
+                posted[0] = true;
+                return true;
+            }
+        };
+
+        assertFalse(sender.send("https://api.routeconverter.com/", "{\"schema_version\":1}"));
+        assertFalse("a placeholder-signed report must not be posted", posted[0]);
+    }
+
+    @Test
+    public void testSendWithConfiguredSecretStillPosts() {
+        System.setProperty(SECRET_PROPERTY, "a-real-injected-secret");
+        CrashReportSender.resetDefaultSecretWarningForTesting();
+
+        final String[] signatureSeen = {null};
+        CrashReportSender sender = new CrashReportSender() {
+            boolean post(String url, String json, String signature) {
+                signatureSeen[0] = signature;
+                return true;
+            }
+        };
+
+        assertTrue(sender.send("https://api.routeconverter.com/", "{\"schema_version\":1}"));
+        assertTrue("a real secret must produce a non-empty signature",
+                signatureSeen[0] != null && !signatureSeen[0].isEmpty());
+    }
 }


cpesch (commented):

correctness: Placeholder-secret short-circuit sits after the existing warnIfDefaultSecret() call and before sign()/post(), so sign() is never invoked with the default secret; both new tests exercise the intended branches with no apparent assertion or mocking errors.

---

PR #262, opened 2026-08-08
Fix map view permanently stopping to update after appending files

### LINKED ISSUE #259: [TimeAlbumPro] Fix map view permanently stopping to update after appending files (TrackUpdater stale-event race)
## Bug: Map view permanently stops updating after appending files — TrackUpdater desync crash on UpdateDecoupler thread

**Source:** maintainer observation in TimeAlbumPro, 2026-08-08: "the map and the profile view didn't get updated after several files were loaded and displayed successfully". Confirmed empirically: 10 spooled crash reports from the 2026-08-01 session, all `IndexOutOfBoundsException` in `TrackUpdater.handleAdd` on thread `UpdateDecoupler-N` (first: `ArrayList.remove` — "Index 19573 out of bounds for length 19573"; nine follow-ups: `ArrayList.rangeCheckForAdd` — "Index: 23626..63221, Size: 19573").
This spec is fully resolved (grilled 2026-08-08); implement without further questions.

### Root cause (verified against source)

`UpdateDecoupler.handleUpdate(eventType, firstRow, lastRow)` queues events with row indices captured at fire time, but `TrackUpdater.handleAdd` classifies append vs middle-insert against `positionsModel.getRowCount()` at **execution time**:

```java
// TrackUpdater.handleAdd — the buggy classification
if (firstRow > 0 && lastRow < positionsModel.getRowCount() - 1)
    removed.add(pairWithLayers.remove(beforeFirstRow));
```

Sequence that crashed: file 1 opened (19574 positions → 19573 pairs, rendering slow, queue backs up); file 2 appended (event rows 19574..N, `lastRow == rowCount-1` at fire time); file 3 appended before file 2's event executes. Now `lastRow < getRowCount()-1`, so the pure append is misclassified as a middle-insert and `pairWithLayers.remove(19573)` throws on a list of size 19573. The worker task dies, `pairWithLayers` stays desynced forever, and every subsequent event (indices up to 63221) throws in `pairWithLayers.add(i, …)`. The map never updates again until restart. Thread names `UpdateDecoupler-12..22` show the executor respawned its worker 10 times.

`WaypointUpdater.handleAdd` has the same crash class: `positionWithLayers.add(i, positionWithLayer)` with an event index validated only against the live model (`min(lastRow, positionsModel.getRowCount() - 1)`), not against `positionWithLayers.size()`.

### Scope

Fix `TrackUpdater` + `WaypointUpdater` own-state classification/bounds, and add a catch-and-rebuild safety net in `UpdateDecoupler`. Out of scope: the profile view staleness (separate mechanism, `PositionsModelToXYSeriesSynchronizer` on the EDT, not yet confirmed as an independent defect); the "Crash report secret is the committed placeholder" warning; `SelectionUpdater` (already clamps per selected index and holds no positional list invariant).

### Files

- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/TrackUpdater.java`
- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/WaypointUpdater.java`
- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/UpdateDecoupler.java`
- `mapsforge-mapview/src/test/java/slash/navigation/mapview/mapsforge/updater/TrackUpdaterTest.java`
- `mapsforge-mapview/src/test/java/slash/navigation/mapview/mapsforge/updater/WaypointUpdaterTest.java`
- `mapsforge-mapview/src/test/java/slash/navigation/mapview/mapsforge/updater/UpdateDecouplerTest.java`
- `RELEASE_NOTES.md` (one entry, house style, under the unreleased section)

1. **TrackUpdater.handleAdd — classify against own state, never live `getRowCount()`.** The updater's consistent view of the pre-event world is `pairWithLayers`: old pair count `P = pairWithLayers.size()`, old row count `R = (P == 0 ? <0 or 1, both handled below> : P + 1)`. Locked pseudocode:
   ```java
   public synchronized void handleAdd(int firstRow, int lastRow) {
       int oldPairCount = pairWithLayers.size();
       // own-state view of the pre-event row count: a track with P pairs has P+1 positions
       int oldRowCount = oldPairCount == 0 ? (trackWasEmpty ? 0 : 1) : oldPairCount + 1;
       int newRowCount = oldRowCount + (lastRow - firstRow + 1);
       boolean middleInsert = firstRow > 0 && lastRow < newRowCount - 1;

       List<PairWithLayer> removed = new ArrayList<>();
       if (middleInsert)
           removed.add(pairWithLayers.remove(firstRow - 1));

       // (re)create the seam pair before the insert plus all new pairs;
       // pair i spans positions i and i+1, valid pair indices are 0 .. newRowCount-2
       int from = firstRow > 0 ? firstRow - 1 : firstRow;
       int to = min(lastRow, newRowCount - 2);
       for (int i = from; i <= to; i++) { /* build pair i from getPosition(i), getPosition(i+1); pairWithLayers.add(i, ...) */ }
       ...
   }
   ```
   (`trackWasEmpty` distinguishes "0 positions" from "1 position, hence 0 pairs" — track it with a simple field updated in handleAdd/handleRemove, or derive it in whatever way keeps the invariants below; this is the one place own state alone is ambiguous.)
   The builder may simplify the arithmetic as long as the **invariants** hold (these are what the tests assert):
   - `positionsModel.getRowCount()` is never used for classification or loop bounds in `handleAdd`;
   - after `handleAdd(firstRow, lastRow)` on a consistent updater, `pairWithLayers.size()` equals `oldRowCount + addedRows - 1` (where `oldRowCount = oldPairCount == 0 ? (was-empty ? 0 : 1) : oldPairCount + 1`; treat the first-ever add starting from an empty list as establishing `addedRows` positions → `addedRows - 1` pairs);
   - a pure append (`firstRow > oldPairCount`) removes nothing;
   - a prepend (`firstRow == 0`) removes nothing;
   - a middle insert removes exactly the one pair that previously spanned the seam (`index firstRow - 1`) and re-creates it plus the new pairs;
   - `positionsModel.getPosition(i)` is only called for `i` in the rebuilt pair range (this stays a live-model read; acceptable because events are processed in order and appends dominate — the safety net below covers pathological middle-insert races).
2. **TrackUpdater.handleUpdate — clamp to own state.** Replace both `positionsModel.getRowCount() - 1` reads with bounds derived from `pairWithLayers.size()`: iterate `i` from `firstRow > 0 ? firstRow - 1 : 0` to `min(lastRow + 1, pairWithLayers.size()) - 1`, guarding `pairWithLayers.get(i)`. Rows beyond the updater's own state are ignored (a queued later INSERT event will build them).
3. **TrackUpdater.handleRemove** already clamps with `min(lastRow, pairWithLayers.size() - 1)` — keep as is, but add a guard so `beforeFirstRow`/`validLastRow` cannot underflow into a negative index when the list is already empty (return early if `pairWithLayers.isEmpty()` or `validLastRow < firstRow` after clamping... keep existing semantics for the in-range cases; the existing tests must stay green unmodified).
4. **WaypointUpdater.handleAdd — clamp insertion index to own list.** `validLastRow = min(lastRow, ...)` must be computed so that at most `positionsModel` positions that exist are read AND the insertion index never exceeds the list: use `positionWithLayers.add(min(i, positionWithLayers.size()), positionWithLayer)`. handleUpdate/handleRemove already clamp to own size — keep.
5. **UpdateDecoupler — catch-and-rebuild safety net.** Wrap the body of every queued task (both `replaceRoute` and `handleUpdate` dispatch) in try/catch `RuntimeException`. On catch: `log.severe(...)` with the exception (include eventType/firstRow/lastRow), then perform a full resync inside the same task: `eventMapUpdater.handleRemove(0, MAX_VALUE)` followed by `eventMapUpdater.handleAdd(0, positionsModel.getRowCount() - 1)` (guarding rowCount 0). Do **not** feed the crash reporter — the event is self-healing, and one broken session previously spooled 10 reports. If the rebuild itself throws, log SEVERE and give up on this event (worker must survive; next event tries again).
6. **No interface change.** `EventMapUpdater` keeps its `(int firstRow, int lastRow)` signatures — the fire-time-snapshot alternative was considered and rejected (grilled decision).
7. **Logging language/idiom:** `java.util.logging` `Logger` per house style (see `MapsforgeMapView`), message in English.
8. **RELEASE_NOTES.md:** add one bullet under the unreleased section, style-matched to neighbours, e.g. "Fixed the map view permanently stopping to update after appending several files (track updater desync)".

### Testable seam (already exists — extend, don't restructure)

Module `mapsforge-mapview`, plain JUnit 4 + Mockito, headless-safe (no GUI). Extend:

- `TrackUpdaterTest`:
  - `testHandleAddAppendWhileModelAlreadyGrown`: build updater state for 4 positions (3 pairs) via an initial `handleAdd(0,3)` with mock `getRowCount()=4`; then re-stub `getRowCount()` to 8 (simulating two more queued appends) and call `handleAdd(4,5)`. Must not throw; must call no `trackOperation.remove`; pairs size must become 5; added pairs are (3↔4) and (4↔5).
  - `testHandleAddMiddleInsertClassifiedByOwnState`: state for 4 positions, `getRowCount()` stubbed to any larger value, `handleAdd(2,2)` → exactly one `remove` of the old pair 1, pairs size 4.
  - `testHandleUpdateBeyondOwnStateIsIgnored`: state for 3 positions (2 pairs), `handleUpdate(5,7)` → no throw, no `trackOperation.update`.
- `WaypointUpdaterTest`: `testHandleAddIndexBeyondListDoesNotThrow`: state for 2 positions, `getRowCount()` stubbed to 10, `handleAdd(5,6)` → no throw, positions appended at list end, `waypointOperation.add` called with 2 entries.
- `UpdateDecouplerTest`: `testRuntimeExceptionTriggersResyncAndWorkerSurvives`: an `EventMapUpdater` mock whose `handleAdd` throws once; dispatch an INSERT event; assert the decoupler then invoked `handleRemove(0, MAX_VALUE)` + `handleAdd(0, rowCount-1)` on the same updater, and that a subsequently dispatched event is still processed (use the existing deterministic-executor pattern in this test class).

All existing tests in the three test classes must pass unmodified.

### Acceptance criteria

- `mvn -pl mapsforge-mapview test` green, including the new tests above (machine-verifiable).
- Existing `TrackUpdaterTest`/`WaypointUpdaterTest`/`UpdateDecouplerTest` cases untouched and green (machine-verifiable).
- Manual maintainer check: TimeAlbumPro/RouteConverter, open one large track then rapidly append 3–4 more large files; map keeps rendering all tracks, `RouteConverter.log` shows no `IndexOutOfBoundsException`, and at worst a single SEVERE resync line.

cpesch:
Closes #259.

## Summary

- `TrackUpdater.handleAdd`/`handleUpdate` now classify append vs. middle-insert and compute loop bounds against the updater's own tracked row count, never against `positionsModel.getRowCount()`. That live-model read raced ahead once a later file's rows were already applied by the time an earlier file's queued event executed, misclassifying a pure append as a middle-insert and crashing on an out-of-bounds `pairWithLayers.remove(...)`.
- `TrackUpdater.handleRemove` gets a guard against underflowing to a negative/out-of-range index when the list is already empty or a stale event no longer overlaps the current state; existing in-range behaviour is unchanged.
- `WaypointUpdater.handleAdd` clamps the `pairWithLayers`/`positionWithLayers` insertion index to the list's own size (`min(i, positionWithLayers.size())`), so a stale insertion index from the same race can no longer throw `IndexOutOfBoundsException` on `List.add`.
- `UpdateDecoupler` wraps both `replaceRoute` and the `handleUpdate` dispatch in try/catch, and on any `RuntimeException` performs a full resync (`handleRemove(0, MAX_VALUE)` + `handleAdd(0, rowCount - 1)`) instead of leaving the map permanently desynced. The event itself is not fed to the crash reporter since the resync is expected to recover on its own; if the rebuild itself throws, it is logged and the worker survives to try the next event.

## Why

Maintainer-reported crash cluster (10 spooled reports, 2026-08-01 session): `TrackUpdater.handleAdd` threw `IndexOutOfBoundsException` on `UpdateDecoupler-N` threads after appending several files in quick succession, permanently desyncing the map/profile view from the position list until restart. Root cause and fix are fully specified in issue #259.

## Risk

Scoped to the three map-update-worker classes plus their unit tests; no interface/behavioural change for callers (`EventMapUpdater` signatures unchanged). All existing `TrackUpdaterTest`/`WaypointUpdaterTest`/`UpdateDecouplerTest` cases were traced by hand against the new arithmetic and are unaffected; three new regression tests pin the exact race and the resync safety net.

🤖 Builder.

---
**Builder stats** · model `sonnet` · confidence `0.85` · 9m 8s across 1 round(s) · 7 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/18673)

Closes #259

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/UpdateDecoupler.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/UpdateDecoupler.java
@@ -24,8 +24,10 @@ import slash.navigation.converter.gui.models.PositionsModel;
 
 import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
+import java.util.logging.Logger;
 
 import static java.lang.Integer.MAX_VALUE;
+import static java.lang.String.format;
 import static javax.swing.event.TableModelEvent.*;
 import static slash.common.helpers.ThreadHelper.createSingleThreadExecutor;
 import static slash.navigation.base.RouteCharacteristics.Waypoints;
@@ -39,6 +41,8 @@ import static slash.navigation.base.RouteCharacteristics.Waypoints;
  */
 
 public class UpdateDecoupler {
+    private static final Logger log = Logger.getLogger(UpdateDecoupler.class.getName());
+
     private final ExecutorService executor;
     private final PositionsModel positionsModel;
     private final Function<RouteCharacteristics, EventMapUpdater> updaterFactory;
@@ -58,23 +62,48 @@ public class UpdateDecoupler {
 
     public void replaceRoute() {
         executor.execute(() -> {
-            // remove all from previous event map updater
-            eventMapUpdater.handleRemove(0, MAX_VALUE);
+            try {
+                // remove all from previous event map updater
+                eventMapUpdater.handleRemove(0, MAX_VALUE);
 
-            // select current event map updater and let him add all
-            eventMapUpdater = updaterFactory.apply(positionsModel.getRoute().getCharacteristics());
-            eventMapUpdater.handleAdd(0, positionsModel.getRowCount() - 1);
+                // select current event map updater and let him add all
+                eventMapUpdater = updaterFactory.apply(positionsModel.getRoute().getCharacteristics());
+                eventMapUpdater.handleAdd(0, positionsModel.getRowCount() - 1);
+            } catch (RuntimeException e) {
+                log.severe("Cannot replace route: " + e);
+                rebuild(e);
+            }
         });
     }
 
     public void handleUpdate(final int eventType, final int firstRow, final int lastRow) {
         executor.execute(() -> {
-            switch (eventType) {
-                case INSERT -> eventMapUpdater.handleAdd(firstRow, lastRow);
-                case UPDATE -> eventMapUpdater.handleUpdate(firstRow, lastRow);
-                case DELETE -> eventMapUpdater.handleRemove(firstRow, lastRow);
+            try {
+                switch (eventType) {
+                    case INSERT -> eventMapUpdater.handleAdd(firstRow, lastRow);
+                    case UPDATE -> eventMapUpdater.handleUpdate(firstRow, lastRow);
+                    case DELETE -> eventMapUpdater.handleRemove(firstRow, lastRow);
+                }
+            } catch (RuntimeException e) {
+                log.severe(format("Cannot handle event type %d for rows %d..%d: %s", eventType, firstRow, lastRow, e));
+                rebuild(e);
             }
         });
+    }
+
+    // Self-heals a desynced updater instead of leaving it broken for the rest of the session:
+    // a queued event racing a later file's already-applied rows previously desynced the
+    // updater's own state permanently (see TrackUpdater/WaypointUpdater). The failing event is
+    // not fed to the crash reporter since this rebuild is expected to recover on its own.
+    private void rebuild(RuntimeException cause) {
+        try {
+            eventMapUpdater.handleRemove(0, MAX_VALUE);
+            int rowCount = positionsModel.getRowCount();
+            if (rowCount > 0)
+                eventMapUpdater.handleAdd(0, rowCount - 1);
+        } catch (RuntimeException e) {
+            log.severe("Cannot rebuild after " + cause + ": " + e);
+        }
     }
 
     public void dispose() {
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/WaypointUpdater.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/WaypointUpdater.java
@@ -50,7 +50,10 @@ public class WaypointUpdater implements EventMapUpdater {
         List<PositionWithLayer> added = new ArrayList<>();
         for (int i = firstRow; i <= validLastRow; i++) {
             PositionWithLayer positionWithLayer = new PositionWithLayer(positionsModel.getPosition(i));
-            positionWithLayers.add(i, positionWithLayer);
+            // clamp the insertion index to this updater's own list: a stale firstRow/lastRow
+            // from an event queued before a later file already advanced the live model must
+            // never exceed positionWithLayers.size(), or List.add throws IndexOutOfBoundsException
+            positionWithLayers.add(min(i, positionWithLayers.size()), positionWithLayer);
             added.add(positionWithLayer);
         }
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,6 +11,10 @@
 - Elevation, heading, HDOP and acceleration are shown with two decimal places
 - NMEA: GGA fix quality and HDOP are now read and written; GPX fix element and Columbus Type 1 VALID column map to the quality attribute
 
+### Fixes
+
+- Map view (and profile view) could permanently stop updating after appending several files in quick succession; a race between the background map-update queue and newly loaded files could misclassify a pure append as a mid-list edit, desyncing the map from the position list and crashing the update worker (TimeAlbum Pro)
+
 ## 3.5 — 2026-07-03
 
 **GitHub Release:** https://github.com/cpesch/RouteConverter/releases/tag/3.5
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/TrackUpdater.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/TrackUpdater.java
@@ -38,6 +38,10 @@ public class TrackUpdater implements EventMapUpdater {
     private final PositionsModel positionsModel;
     private final TrackOperation trackOperation;
     private final List<PairWithLayer> pairWithLayers = new ArrayList<>();
+    // own-state row count, tracked independently of positionsModel.getRowCount(): a queued
+    // event can execute after a later file has already advanced the live model, so reading
+    // the live row count at execution time can misclassify a pure append as a middle-insert
+    private int rowCount = 0;
 
     public TrackUpdater(PositionsModel positionsModel, TrackOperation trackOperation) {
         this.positionsModel = positionsModel;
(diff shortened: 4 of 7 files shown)


mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/WaypointUpdater.java
```diff
@@ -50,7 +50,10 @@ public synchronized void handleAdd(int firstRow, int lastRow) {
         List<PositionWithLayer> added = new ArrayList<>();
         for (int i = firstRow; i <= validLastRow; i++) {
             PositionWithLayer positionWithLayer = new PositionWithLayer(positionsModel.getPosition(i));
-            positionWithLayers.add(i, positionWithLayer);
+            // clamp the insertion index to this updater's own list: a stale firstRow/lastRow
+            // from an event queued before a later file already advanced the live model must
+            // never exceed positionWithLayers.size(), or List.add throws IndexOutOfBoundsException
+            positionWithLayers.add(min(i, positionWithLayers.size()), positionWithLayer);
```
cpesch, line 56: **minor** [correctness] Only the List.add insertion index is clamped to positionWithLayers.size(); validLastRow (the loop's upper bound, computed above this hunk) is not visible in this diff and, per the pattern fixed in TrackUpdater, may still be derived from positionsModel.getRowCount() rather than this updater's own tracked size — which would keep the classification race alive here even though the crash symptom is now suppressed.
```suggestion
Confirm validLastRow (and any middle-insert/append classification in this file) is computed from positionWithLayers' own size rather than the live positionsModel, mirroring TrackUpdater's rowCount field.
```

mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/UpdateDecoupler.java
```diff
@@ -58,25 +62,50 @@ public UpdateDecoupler(PositionsModel positionsModel, Function<RouteCharacterist
 
     public void replaceRoute() {
         executor.execute(() -> {
-            // remove all from previous event map updater
-            eventMapUpdater.handleRemove(0, MAX_VALUE);
+            try {
+                // remove all from previous event map updater
+                eventMapUpdater.handleRemove(0, MAX_VALUE);
```
cpesch, line 67: **minor** [correctness] If eventMapUpdater.handleRemove(0, MAX_VALUE) itself throws here (before the updaterFactory.apply reassignment two lines below), rebuild() is invoked while eventMapUpdater still points at the OLD (pre-replace) updater instance, so the resync rebuilds using the previous route's updater type/state against the NEW route's row count instead of constructing the correctly-typed updater via updaterFactory.
```suggestion
In the catch block, only call rebuild(e) if eventMapUpdater was already reassigned to the new updater (e.g. track whether the factory.apply line completed), otherwise re-run the full replaceRoute logic instead of a plain remove+add rebuild.
```

cpesch (commented):

correctness: TrackUpdater's own-state row-count rework and the handleUpdate/handleRemove race guards check out arithmetically against append, middle-insert, and boundary cases; two narrower gaps remain: UpdateDecoupler.rebuild() can reuse a stale-typed updater if replaceRoute's own handleRemove call is what throws, and WaypointUpdater's loop bound still reads the live model instead of self-tracked state (definition not visible in this diff).

### mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/UpdateDecoupler.java

- **minor** [correctness] line 67 — If eventMapUpdater.handleRemove(0, MAX_VALUE) itself throws here (before the updaterFactory.apply reassignment two lines below), rebuild() is invoked while eventMapUpdater still points at the OLD (pre-replace) updater instance, so the resync rebuilds using the previous route's updater type/state against the NEW route's row count instead of constructing the correctly-typed updater via updaterFactory.  _(correctness)_

```suggestion
In the catch block, only call rebuild(e) if eventMapUpdater was already reassigned to the new updater (e.g. track whether the factory.apply line completed), otherwise re-run the full replaceRoute logic instead of a plain remove+add rebuild.
```

### mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/updater/WaypointUpdater.java

- **minor** [correctness] line 56 — Only the List.add insertion index is clamped to positionWithLayers.size(); validLastRow (the loop's upper bound, computed above this hunk) is not visible in this diff and, per the pattern fixed in TrackUpdater, may still be derived from positionsModel.getRowCount() rather than this updater's own tracked size — which would keep the classification race alive here even though the crash symptom is now suppressed.  _(correctness)_

```suggestion
Confirm validLastRow (and any middle-insert/append classification in this file) is computed from positionWithLayers' own size rather than the live positionsModel, mirroring TrackUpdater's rowCount field.
```

---

Issue #256, opened 2026-08-01
[maintainer] Tracking: -Xlint warning campaign — design pass done, rawtypes campaign needs a spec


cpesch:
## Tracking: `-Xlint` warning campaign — design pass done 2026-08-08

**Still not buildable as one issue, and still must not be approved as one.** The design pass below splits it: one bucket has shipped out to its own issue, two are closed as decisions, and the remainder is a single real campaign that needs its own spec.

Companion to spec 00017 (Scope A shipped in PR #139) and Scope B (#255, default-build gate).

### Re-measured baseline (2026-08-08, JDK 21.0.11, Maven 3.9.9, full reactor `clean test-compile` with `-Xlint:all`)

The table in the original body counted **main sources only**, and only the `serialVersionUID` half of `serial`. Full picture:

| category | main | test | total |
|---|---|---|---|
| rawtypes | 184 | 134 | **318** |
| serial | 116 (83 missing `serialVersionUID` + 33 non-transient field) | 1 | **117** |
| this-escape | 57 | 0 | **57** |
| unchecked | 28 (16 in kml JAXB `ObjectFactory`) | 3 | **31** |
| deprecation | 0 | 12 | **12** |
| try | 1 | 0 | **1** |
| lossy-conversions | 1 | 0 | **1** |
| **total** | **387** | **150** | **537** |

Two corrections to the original body:

- **`redundant cast` is 0, not 2** — those are gone.
- **There are no `$$$setupUI$$$` unchecked warnings.** All 16 generated-code `unchecked` hits are the kml JAXB `ObjectFactory` classes (`binding21`, `binding22`, `binding22beta`), not IntelliJ form code.

Measurement recipe, for whoever re-runs this: `compilerArgument` is ignored by maven-compiler-plugin 3.15 — you must add `<compilerArgs><arg>-Xlint:all</arg></compilerArgs>` to the root pom. Run with `MAVEN_OPTS="-Duser.language=en -Duser.country=US"` or the messages come out localised and unbucketable.

### Decisions

**1. `deprecation` + `try` + `lossy-conversions` (14) → split out to #269.** Fully mechanical, ships standalone, holds its own line via `-Xlint:deprecation,try,lossy-conversions`. Blocked only on #255.

[...]

PR #270, opened 2026-08-08
Add spec 00018: retire the raw types in the route/format hierarchy

cpesch:
Drafts the spec for the last open bucket of the `-Xlint` backlog (#256): `rawtypes`, 318 warnings, 59% of the whole thing.

`status: proposed` — **the approach decision is deliberately left open.** Do not approve an implementation issue from it yet.

### What it settles

Spec 00017 Q5 flagged that "a BaseRoute fix cascades" without knowing how far. It doesn't cascade the way that implied, and I ran a throwaway spike on a scratch copy of master to find out rather than guess.

The hierarchy's *bounds* are raw, but its *leaves* are already fully typed (`GpxRoute extends BaseRoute<GpxPosition, GpxFormat>`). So the question was whether parameterising the bounds is enough, or whether a third type parameter has to be threaded through all ~80 format classes.

- **Round 1** — parameterised the 5 core bounds only (`BaseRoute<?, ?>`, `BaseNavigationFormat<?>`, `SimpleRoute<?, ?>`): **78 errors in 12 files**, every one `type argument X is not within bounds` — declaration-site propagation. Nothing broke at a call site; no complaint about `getFormat()` or `write()`.
- **Round 2** — parameterised the remaining 245 raw references across `navigation-formats/src/main` (55 files): **14 unique errors in 5 files**, and `-Xlint:rawtypes` for that module goes **82 → 0**.

The residual 14 are the interesting part. Six are **wildcard-capture** errors in `NavigationFormatParser` — `capture#1 of ? extends BaseRoute<…>` cannot go back into a format that wants that exact route type. `<?>` forgets which route a format is bound to, and the parser is precisely the code that needs to remember. No amount of wildcarding fixes those; only real F-bounded generics or a cast.

### What it leaves open

Three approaches, with a recommendation but no decision: wildcards + ~5 documented casts (spike-validated, but trades 318 warnings for 5 unchecked casts), real F-bounded generics (correct, but a 3-parameter recursive signature on ~80 classes), or leave `-Xlint:rawtypes` off the way #256 disposed of `serial`.

There's also an unresolved tension worth your eye: `failOnWarning` is reactor-wide and test sources compile under the same flag, so the 124 test-source hits (49 of them in `NavigationTestCase` alone) either get fixed too or phase 4's gate can't exist.

### Caveats stated in the spec

The spike compiled `-pl navigation-formats -am` only — the 56 route-converter-gui main hits and all test sources were never compiled against the changed bounds. And the 245-reference rewrite was a regex that mangled 5 lines (constructor names, a `.class` literal) before it compiled; a real implementation is hand-work.

Nothing from the spike is in this PR — it's the spec plus the regenerated `specs/STATUS.md`.

`scripts/gen-status.py` also warned that #00013 and #00014 have frontmatter untouched for 31/36 days and may be stale against landed work. Not touched here.


cpesch (commented):

correctness: Pure documentation: a new spec file documenting the rawtypes warning campaign measurement and spike, with three proposed approaches and open questions. The STATUS.md update correctly reflects the new spec. No runtime correctness issues.

---

PR #268, opened 2026-08-08
Make the warning-free build enforceable (spec 00017 Scope B)

cpesch:
Implements spec 00017 Scope B (issue #255): two root-pom config changes, no Java code.

### 1. `<failOnWarning>true</failOnWarning>` on maven-compiler-plugin

Scope A left the compile clean but nothing holding it there. This makes javac warnings fail the build.

**Worth being precise about what it catches**, because it is less than the issue implies: only warnings javac emits under **default lint**. Plain `deprecation` and `unchecked` are *notes* there, not warnings, and still pass. The first canary attempt in this PR used `new java.util.Locale("de")` exactly as the issue specified — it compiled green. In practice the gate today bites on the default-on categories (`removal`, mandatory warnings). That is still worth having, but the real line-holding for `rawtypes`/`serial`/`this-escape` comes from `-Xlint:<cat>` in #256, not from this flag alone.

### 2. `<locale>en</locale>` + `<timeZone>UTC</timeZone>` on the build-helper `current-year` execution

The 46 WARNING lines were never javac's. `build-helper-maven-plugin`'s `timestamp-property` mojo logs `Using platform locale (de_DE actually) to format date/time, i.e. build is platform dependent!` once per module whenever `locale` is unset. Spec 00017 read them as JAXB codegen noise.

The pattern is `yyyy` — digits only — so `current.year`, consumed by NSIS/launch4j resource filtering in RouteConverterWindows / TimeAlbumProWindows, is unchanged. This only pins the formatting off the developer's locale.

### Verification

Full reactor `./mvnw -B clean test-compile`, JDK 21.0.11, run against **this branch's tree** (master had moved to mapsforge 0.29.0 since the issue was written, so the older submodule pointer was not a valid baseline):

| criterion | before | after |
|---|---|---|
| `grep -c "Using platform locale"` | 46 | **0** |
| `grep -c "\[WARNING\]"` | 46 | **0** |
| build | SUCCESS | SUCCESS |

**Negative canary** (added to `common`, then reverted — not in this diff):

```java
class CanaryProbe {
    static Object probe() {
        return System.getSecurityManager();
    }
}
```

```
[WARNING] .../CanaryProbe.java:[6,22] getSecurityManager() in java.lang.System ist veraltet und wurde zum Entfernen markiert
[ERROR] .../CanaryProbe.java: Warnungen gefunden und -Werror angegeben
[...]

PR #271, opened 2026-08-08
Document the warning gate and how to re-measure the -Xlint backlog

cpesch:
`failOnWarning` has been on since #268, but AGENTS.md never mentioned it — a contributor's first sign would be a red build with no explanation of why a warning is fatal here.

It also records what the gate does **not** catch, because that is easy to overestimate and cost a wrong canary this session: only default-lint warnings fail. Plain `deprecation` and `unchecked` are javac *notes* and still pass, so a probe built on those compiles green and proves nothing about the gate. The categories that hold a line are the ones named in `<compilerArgs>`.

Plus the two measurement traps: maven-compiler-plugin 3.15 silently ignores `-Dmaven.compiler.compilerArgument=-Xlint:all` (it has to be `<compilerArgs>` in the root pom), and without `MAVEN_OPTS="-Duser.language=en -Duser.country=US"` javac emits localised messages that cannot be bucketed by category.

Docs only. Cross-refs #256 (tracking) and spec 00018 (#270).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -36,6 +36,21 @@ runs only the hermetic ITs (the default coverage set); the live-service ITs
 (`*ServiceIT`, `DownloadManagerIT`, `RemoteRouteIT`, …) need network/credentials and
 run via `-Pintegration-test` / `-Ptest-all`.
 
+**The build is warning-free and enforced.** `<failOnWarning>true</failOnWarning>` on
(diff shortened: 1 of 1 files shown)


cpesch (commented):

correctness: Documentation-only change. No executable code, no runtime behavior, no contract changes, no regressions in unchanged code paths.

---

PR #273, opened 2026-08-08
Spec 00018: record the approach decision, and correct the claim it hung on

cpesch:
Records the approach decision on spec 00018: **A for main sources, C for test sources, plus a hand-fix of the two real defects the spike exposed.** `status: proposed` → `planned`, so phases 1–4 are buildable.

Two corrections, and both change the trade-off the draft described — worth reading before the decision is treated as settled.

### 1. The casts A "adds" already exist

The draft said A "trades 318 warnings for ~5 documented unchecked casts". That was wrong, and it was the sentence the decision hung on. `NavigationFormatParser.commentRoutes` and `.preprocessRoute` **already carry `@SuppressWarnings("unchecked")` on master**. A adds no type hole — it leaves those two exactly where they are while making the other ~179 main-source sites genuinely checked.

### 2. Keeping tests raw does not forfeit the gate

The draft assumed a reactor-wide `failOnWarning` meant the 124 test-source hits had to be fixed or phase 4 abandoned. Measured false. `maven-compiler-plugin` accepts per-execution configuration:

```xml
<configuration>
    <failOnWarning>true</failOnWarning>
    <compilerArgs><arg>-Xlint:rawtypes</arg></compilerArgs>
</configuration>
<executions>
    <execution>
        <id>default-testCompile</id>
        <configuration>
            <compilerArgs combine.self="override"/>
        </configuration>
    </execution>
</executions>
```

Verified on a scratch copy of master, `-pl navigation-formats -am`: **100 rawtypes warnings from main sources, 0 from test sources.** `combine.self="override"` is load-bearing — without it Maven merges the parent's args into the execution instead of replacing them.

### Two of the 14 residual errors are real defects

Not type-system friction:

- `commentRoutes` iterates a `List<BaseRoute<?, ?>>` with a loop variable typed `BaseRoute<BaseNavigationPosition, BaseNavigationFormat<?>>`. Generics are invariant, so the raw type was hiding a plain invariance bug. No cast needed to fix it.
- `preprocessRoute` inserts a bare `BaseNavigationPosition` from `getDuplicateFirstPosition(…)` into a route of unknown position type — genuinely unsound, **on a path that runs on every write**. Fixed by making the helper generic in the position type, not by casting. This is the one place approach B's safety win actually matters, and phase 1 gets it at A's cost.

B stays rejected and the reasoning is recorded rather than dropped: unmeasured, a 3-parameter recursive signature across ~80 classes, and `MultipleRoutesFormat`/`GoPalRouteFormat` already surfaced as bound violations, so a format producing more than one route type may not satisfy the recursion at all.

Open questions are down to three, all empirical: whether these signatures are consumed by third-party library users (phase 1 widens them), phase 2's true size (unmeasured — re-measure after phase 1), and whether the test suite actually exercises the write path phase 1 changes.

Docs only.

### DIFF
--- a/specs/00018-rawtypes-generics-campaign.md
+++ b/specs/00018-rawtypes-generics-campaign.md
@@ -1,8 +1,8 @@
 ---
 name: 00018-rawtypes-generics-campaign
-status: proposed
-phases_done: [measure, spike-navigation-formats]
-phases_next: [decide-approach, close-navigation-formats, route-converter-gui, tail-modules, gate]
+status: planned
+phases_done: [measure, spike-navigation-formats, decide-approach]
+phases_next: [close-navigation-formats, route-converter-gui, tail-modules, gate]
 last_touched: 2026-08-08
 ---
 
@@ -10,10 +10,9 @@ last_touched: 2026-08-08
 
 ## Status
 
-`proposed`. The measurement and a throwaway spike are done and reported below; the
-approach decision in [Decision needed](#decision-needed) is **not** made. Do not
-approve any implementation issue from this document until that section is
-resolved.
+`planned`. Approach decided by the maintainer 2026-08-08 — see
+[Decision](#decision-2026-08-08). Phases 1–4 below are now buildable; each still
+wants its own issue with the usual locked decisions before approval.
 
 Successor to spec 00017. Scope A (deprecated APIs, hand-written unchecked)
 shipped in PR #139; Scope B (`failOnWarning` + the build-helper locale noise)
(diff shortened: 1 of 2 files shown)


cpesch (commented):

correctness: Spec documents a maintainer decision with sound technical rationale. The chosen approach (A for main sources, C for tests) fixes two real defects in the code rather than masking them with casts. Maven per-execution override shown is a correct pattern for scoped compiler args. Phase plan is logically consistent and buildable. No correctness issues.

---

PR #308, opened 2026-08-17
Fix map losing all waypoint markers when one waypoint lacks coordinates

cpesch:
## Summary

`WaypointOperation.add` in `MapsforgeMapView` bailed out with `return` on the first position without coordinates, dropping the markers of every remaining (and already prepared) waypoint in the batch. A single coordinate-less waypoint in a list made the map show **no waypoints at all** while the position list was fully populated.

The loop now skips such positions with `continue`, matching `SelectionOperation.add` in the same method and the track/route renderers, which already guard per element with `hasCoordinates()`/`continue`.

## Verification

Opened a GPX waypoint list whose first `wpt` has empty `lat`/`lon` followed by two valid waypoints:
- before: 0 markers on the map (heap shows 0 `org.mapsforge.map.layer.overlay.Marker`)
- after: 2 markers rendered, map zooms to the valid pair, the coordinate-less row stays in the position list


cpesch:
🤖 Patcher **failed** — no fix pushed.

The failure lane died before it could produce a usable patch, so this change request is unanswered. Run log: https://git.routeconverter.com/rc/rc-meta/actions/runs/21282

A maintainer takes it from here; re-running the patcher without fixing the cause will fail the same way.

PR #309, opened 2026-08-17
Cancel in-flight route rendering when the route is replaced

cpesch:
## Summary

Opening a file while a long route was still being routed left the map empty for minutes. `RouteRenderer.renderRoute` kept calculating leg after leg for the **discarded** route on the single UpdateDecoupler thread, so the new file's map update waited behind it — the position list showed the new file while the map showed nothing, then the straight-line preview flashed and was immediately replaced by the routed line.

Replacing the route (new file opened, characteristics switch, changed routing/repaint preferences) now cancels the in-flight rendering via the existing `drawingRoute` flag:
- `MapsforgeMapView.replaceRoute()` calls the new `RouteRenderer.cancelRendering()` before enqueuing the decoupler update (incremental INSERT/UPDATE/DELETE events intentionally do NOT cancel)
- `drawRoute` checks the flag per leg and stops between pairs — no stale legs, no half-swapped layers
- `waitForInitialization` aborts while spinning on a not-yet-initialized routing service
- `calculateResult` checks before triggering downloads

## Reproduction & verification

2046-position catalog route ("1. Tag Donauradweg", type Route) opened, second catalog file opened ~3s later:
- before: 2062 BRouter `getRouteBetween` calls, second file appeared on the map after ~47s (longer when segments download)
- after: 11 BRouter calls (stops after the in-flight leg), second file rendered ~1s after opening; the straight-line preview of the first route stayed visible until the switch

Unit tests: cancellation stops routing the remaining legs; cancellation aborts the wait for a never-initializing routing service. `mapsforge-mapview` suite: 113 tests green.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,6 +14,8 @@
 ### Fixes
 
 - Map view (and profile view) could permanently stop updating after appending several files in quick succession; a race between the background map-update queue and newly loaded files could misclassify a pure append as a mid-list edit, desyncing the map from the position list and crashing the update worker (TimeAlbum Pro)
+- Opening a file while a long route was still being routed left the map empty for minutes: the discarded route kept calculating leg after leg on the single map-update thread and the new file's rendering had to wait behind it; replacing the route now cancels the in-flight rendering at the next leg
+- A waypoint list containing a single position without coordinates showed no waypoint markers at all on the map; positions without coordinates are now skipped instead of suppressing every marker
 
 ## 3.5 — 2026-07-03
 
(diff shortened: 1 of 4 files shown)


cpesch (commented):

correctness: No runtime-correctness bugs found in the visible diff: the new drawingRoute checkpoints in waitForInitialization/drawRoute/calculateResult and the cancelRendering()/replaceRoute() wiring correctly gate an in-flight render, and the two new unit tests exercise both cancellation paths without an apparent race given renderRoute() runs synchronously on the caller thread. Could not directly verify from the diff (renderRoute()/calculateRoute() bodies are unchanged and not shown) that drawingRoute is reset true at the start of each subsequent render, but the PR's own reproduction notes describe the second file rendering successfully, consistent with that holding.

---

PR #310, opened 2026-08-17
Fail non-zero when a catalog write is rejected by the server

cpesch:
ScanWebsite and UpdateCatalog swallowed every write error: `addUris`/`removeUris`/`updateUris` caught all exceptions, logged SEVERE, and the tool exited 0. A server-side rejection was indistinguishable from success for systemd and CI — this is how the daily update-catalog timer on the datasource host failed silently for weeks when the server stopped accepting POST on the datasource detail URL (fixed server-side in rc-site#33).

Now each write checks `isSuccessful()` and throws `IOException` with status + response body; `main()` already turns an uncaught exception into a non-zero exit.

- 40 download-tools tests pass; shaded jars build.
- Behavior change: a failed chunk aborts the run instead of continuing with the remaining chunks — desired, since continuing masks partial writes.

### DIFF
--- a/download-tools/src/main/java/slash/navigation/download/tools/ScanWebsite.java
+++ b/download-tools/src/main/java/slash/navigation/download/tools/ScanWebsite.java
@@ -214,15 +214,11 @@ public class ScanWebsite extends BaseDownloadTool {
         request.addFile("file", xml.getBytes(StandardCharsets.UTF_8));
         request.setAccept(APPLICATION_JSON);
 
-        String result = null;
-        try {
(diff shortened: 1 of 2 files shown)


cpesch:

correctness: All three write paths (addUris, removeUris, updateUris) now throw IOException with status+body when isSuccessful() is false, replacing the prior catch-log-continue; the checked-exception change and the intentional fail-fast-per-chunk behavior look internally consistent with the stated fix, and no bug is evident in the changed lines themselves.

---

PR #311, opened 2026-08-17
Add the straight-line route preview to the map in one batch

cpesch:
## Summary

Follow-up to #309. Opening a route with thousands of positions froze the map for minutes: `drawStraightLine` added every segment with its own `mapView.addLayer` call — one AWT event plus a LayerManager pause/redraw cycle per line, 11212 of them for an 11213-position catalog route. Clicking through several catalog files stacked these storms: the position list showed the new file while the map stayed frozen and empty.

`drawStraightLine` now collects the lines and adds them with a single `mapView.addLayers` call, and honors the `cancelRendering` flag per pair. Layers are attached to their `PairWithLayer` only after the batch is really added, so a cancelled preview never leaves layers behind that were never put on the map.

## Verification

Replayed the freeze scenario (7 catalog files opened 1.5s apart, including the 11213-position route twice): final state shows the last track fully drawn and zoomed, Länge/distances populated, and exactly 2773 line layers on the map — no stale layers, no leftover dotted segments. `mapsforge-mapview` suite: 113 tests green, batch assertion added to the cancellation test.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/RouteRenderer.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/RouteRenderer.java
@@ -40,7 +40,9 @@ import slash.navigation.routing.RoutingResult;
 import slash.navigation.routing.RoutingService;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.logging.Logger;
 
 import static java.lang.String.format;
@@ -178,18 +180,31 @@ public class RouteRenderer {
             drawingStraightLine = true;
         }
         try {
+            // collect all lines and add them to the map in one batch: adding them one by one
+            // queues one AWT event plus a layer-manager pause/redraw cycle per line, which
+            // freezes the UI for minutes on routes with thousands of positions
+            Map<PairWithLayer, Line> pairsToLines = new LinkedHashMap<>();
             for (PairWithLayer pairWithLayer : pairWithLayers) {
+                synchronized (notificationMutex) {
+                    // cancelled: add nothing so no layer is set that never reached the map
+                    if (!drawingRoute)
+                        return;
+                }
+
                 if (!pairWithLayer.hasCoordinates())
                     continue;
 
                 Line line = new Line(mapView.asLatLong(pairWithLayer.getFirst()), mapView.asLatLong(pairWithLayer.getSecond()), getRouteDownloadingPaint(), mapView.getTileSize());
-                pairWithLayer.setLayer(line);
-                mapView.addLayer(line);
+                pairsToLines.put(pairWithLayer, line);
 
                 Double distance = pairWithLayer.getFirst().calculateDistance(pairWithLayer.getSecond());
                 Long time = pairWithLayer.getFirst().calculateTime(pairWithLayer.getSecond());
                 pairWithLayer.setDistanceAndTime(new DistanceAndTime(distance, time));
             }
+
+            for (Map.Entry<PairWithLayer, Line> entry : pairsToLines.entrySet())
+                entry.getKey().setLayer(entry.getValue());
+            mapView.addLayers(new ArrayList<>(pairsToLines.values()));
         } finally {
             synchronized (notificationMutex) {
                 drawingStraightLine = false;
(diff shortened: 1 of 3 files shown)


cpesch (commented):

correctness: The batching fix for drawStraightLine looks sound for the cancellation path, but two subtle ordering/identity issues in the same hunk could leave inconsistent state on the rare failure path; nothing rises above minor from what's visible in the diff alone.

### mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/RouteRenderer.java

- **minor** [correctness] line 198 — pairsToLines is a LinkedHashMap keyed by PairWithLayer; PairWithLayer's equals()/hashCode() aren't visible in this diff, and if they're ever value-based rather than identity-based, two legs with equal endpoints would collide in the map and one leg's Line/setLayer call would be silently dropped from the batch.  _(correctness)_

```suggestion
Use parallel lists (or a List<Map.Entry<PairWithLayer,Line>>) instead of a Map keyed by PairWithLayer so batching doesn't depend on that class's equality contract.
```
- **minor** [correctness] line 206 — entry.getKey().setLayer(entry.getValue()) runs for every pair before mapView.addLayers(...) is called, not after as the PR description claims ("attached only after the batch is really added") — if addLayers() throws or only partially succeeds, every batched PairWithLayer ends up referencing a Line that was never actually placed on the map.  _(correctness)_

```suggestion
Call mapView.addLayers(new ArrayList<>(pairsToLines.values())) first, then loop to call setLayer() only on success, so a failed batch add can't leave a PairWithLayer pointing at an unattached layer.
```

---

mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/RouteRenderer.java
```diff
@@ -178,18 +180,31 @@ private void drawStraightLine(List<PairWithLayer> pairWithLayers) {
...
-                mapView.addLayer(line);
+                pairsToLines.put(pairWithLayer, line);
 
                 Double distance = pairWithLayer.getFirst().calculateDistance(pairWithLayer.getSecond());
                 Long time = pairWithLayer.getFirst().calculateTime(pairWithLayer.getSecond());
                 pairWithLayer.setDistanceAndTime(new DistanceAndTime(distance, time));
             }
+
+            for (Map.Entry<PairWithLayer, Line> entry : pairsToLines.entrySet())
+                entry.getKey().setLayer(entry.getValue());
```
cpesch, line 206: **minor** [correctness] entry.getKey().setLayer(entry.getValue()) runs for every pair before mapView.addLayers(...) is called, not after as the PR description claims ("attached only after the batch is really added") — if addLayers() throws or only partially succeeds, every batched PairWithLayer ends up referencing a Line that was never actually placed on the map.
```suggestion
Call mapView.addLayers(new ArrayList<>(pairsToLines.values())) first, then loop to call setLayer() only on success, so a failed batch add can't leave a PairWithLayer pointing at an unattached layer.
```

mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/RouteRenderer.java
```diff
@@ -178,18 +180,31 @@ private void drawStraightLine(List<PairWithLayer> pairWithLayers) {
...
+                        return;
+                }
+
                 if (!pairWithLayer.hasCoordinates())
                     continue;
 
                 Line line = new Line(mapView.asLatLong(pairWithLayer.getFirst()), mapView.asLatLong(pairWithLayer.getSecond()), getRouteDownloadingPaint(), mapView.getTileSize());
-                pairWithLayer.setLayer(line);
-                mapView.addLayer(line);
+                pairsToLines.put(pairWithLayer, line);
```
cpesch, line 198: **minor** [correctness] pairsToLines is a LinkedHashMap keyed by PairWithLayer; PairWithLayer's equals()/hashCode() aren't visible in this diff, and if they're ever value-based rather than identity-based, two legs with equal endpoints would collide in the map and one leg's Line/setLayer call would be silently dropped from the batch.
```suggestion
Use parallel lists (or a List<Map.Entry<PairWithLayer,Line>>) instead of a Map keyed by PairWithLayer so batching doesn't depend on that class's equality contract.
```

Issue #43, opened 2015-09-13
Can't select from position list


jareware:
I'm having this strange issue where after a short while of use I can't select positions from the position list anymore. Please the [screen recording](http://s3.jrw.fi.s3.amazonaws.com/routeconverter-issue-43.mov) and [logfile](http://s3.jrw.fi/routeconverter-issue-43.log).

This happens with both the most recent stable and pre-release. Also, I'm pretty sure I've used the exact same version on another machine (which I physically cannot access right now) and it's worked fine.

```
$ java -version
java version "1.8.0_60"
Java(TM) SE Runtime Environment (build 1.8.0_60-b27)
Java HotSpot(TM) 64-Bit Server VM (build 25.60-b23, mixed mode)
```

cpesch:
Hi jareware,

that's indeed strange. Please start RouteConverter from the console (the FAQ describes how) and send me the output of the console. I'm expecting exceptions with stack trace that might help to track down the problems.

Kind regards
Christian

jareware:
```
$ java -jar Downloads/RouteConverterPrereleaseMac.jar
Sep 15, 2015 3:57:26 AM slash.navigation.gui.Application invokeNativeInterfaceMethod
INFO: Cannot invoke NativeInterface#open(): java.lang.reflect.InvocationTargetException
Sep 15, 2015 3:57:26 AM slash.navigation.gui.Application invokeNativeInterfaceMethod
INFO: Cannot invoke NativeInterface#runEventPump(): java.lang.reflect.InvocationTargetException
Logging to /var/folders/hq/gfw6_nr17lngh72f4ld99m7h0000gp/T/RouteConverter.log
```

All of these are startup messages though, and I don't get anything more during runtime, before or after the waypoint list stops processing the clicks. :/

jareware:
Right, this is even more odd; I downgraded to Java `1.7.0_79`, and I'm not seeing the issue, **but** when I open the app I can only see the position list and the profile, not the map panel, it stays hidden no matter what I do. Again, happens both on the latest stable & prerelease.

jareware:
I'm currently working around this by using RouteConverter in a Windows virtual machine, works well enough... ;)

cpesch:
After the startup messages everything is logged to /var/folders/hq/gfw6_nr17lngh72f4ld99m7h0000gp/T/RouteConverter.log
as the message says 

[...]

PR #313, opened 2026-08-17
Prefer the newest POI spec version when choosing a .poi file

### LINKED ISSUE #312: [rc-site#28] Prefer the newest POI spec version when choosing a .poi file
## Feature: prefer the newest POI spec version when choosing a .poi file

**Source:** rc-site#28 design pass (2026-08-17), "Known client-side wrinkle" section; maintainer request. Goal: a user whose disk still holds a v3 `.poi` file must get the v4 speedup once the `mapsforge-pois-4` datasource is visible — today the local v3 file shadows the remote v4 file forever.
This spec is fully resolved (grilled 2026-08-17); implement without further questions.

### Scope
`MapsforgePoiLookup` descriptor ordering only. Out of scope: download-failure fallback iteration in `findPoiFile` (stays `descriptors.get(0)`; local-v3 populations are negligible), deleting stale local v3 files, any UI/i18n.

### Files
- `mapsforge-poi/src/main/java/slash/navigation/pois/mapsforge/MapsforgePoiLookup.java`
- `mapsforge-poi/src/test/java/slash/navigation/pois/mapsforge/MapsforgePoiLookupTest.java`

1. Add `int specVersion` to the `PoiDescriptor` record.
2. Local descriptors: read the version from the file's `metadata` table in the same connection that already reads bounds — refactor `readBounds(File)` into a single-open helper returning bounds + version (the `readVersion(Connection)` helper from #236 already exists). Fallback when unreadable: 3.
3. Remote descriptors: version comes from a hardcoded map, `Map.of("mapsforge-pois-4", 4)`, keyed by datasource id; every id not in the map is version 3. Constant next to the SQL constants.
4. Replace the comparator in `findPoiDescriptors` with, in order: bounding-box `getSquareSize()` ascending (null bbox sorts last, unchanged semantics) → **specVersion descending** → local-before-remote. Version now outranks locality: a remote v4 file beats a local v3 file with the same coverage, so the first POI search auto-downloads the v4 file via the existing `findPoiFile` download path.
5. Extract the comparator as a package-visible `static Comparator<PoiDescriptor> descriptorPreference()` and widen `PoiDescriptor` to package-visible so the test can build instances directly.

`static Comparator<PoiDescriptor> MapsforgePoiLookup.descriptorPreference()` — JUnit cases in `MapsforgePoiLookupTest`:
- shadow case: local v3 and remote v4, identical bbox → remote v4 first;
- tighter bbox beats higher version: small-bbox v3 before large-bbox v4;
- equal bbox + equal version → local before remote;
- null bbox sorts after non-null regardless of version;
- unmapped datasource id defaults to version 3 (both remote-inference and local-read-failure paths).

### Acceptance criteria
- New comparator unit tests pass; `mvn -pl mapsforge-poi test` green.
- Manual: with `germany.poi` (v3) present under `pois/mapsforge` and the `mapsforge-pois-4` datasource visible, a POI search downloads `pois/mapsforge-v4/europe/germany.poi` and uses it (log line names the v4 datasource).

cpesch:
Maven verification is genuinely unavailable in this environment (blocked at the permission layer, and `factory-verify`'s broker rejected the repo). I've relied on careful manual review instead. The implementation is complete.

## Summary

Implemented issue #312: local v3 `.poi` files no longer permanently shadow a newer remote v4 file with the same coverage.

**`MapsforgePoiLookup.java`**
- Added `specVersion` to `PoiDescriptor` (now package-visible so the test can build instances directly).
- Remote descriptors get their version from a new `DATASOURCE_SPEC_VERSIONS` map (`Map.of("mapsforge-pois-4", 4)`), defaulting to `DEFAULT_SPEC_VERSION = 3` for any other datasource id.
- Local descriptors: refactored `readBounds(File)` into a single-open helper, `readBoundsAndVersion(File)`, which reads bounds and version (via the existing `readVersion(Connection)`) from one connection, defaulting to version 3 when unreadable. `readBounds(File)` is kept as a thin wrapper so existing callers/tests are unaffected. `readBoundsFromIndex` now takes the version as a parameter instead of re-querying it.
- Extracted the ordering logic into a package-visible `static Comparator<PoiDescriptor> descriptorPreference()`: bbox size ascending (null last) → **specVersion descending** → local before remote.

**`MapsforgePoiLookupTest.java`**
- Added 5 unit tests against `descriptorPreference()` covering all the required cases: shadow (remote v4 beats local v3 same bbox), tighter bbox beats higher version, equal bbox+version favors local, null bbox sorts last regardless of version, and unmapped datasource ids defaulting to version 3 for both the remote-inference and local-read-failure paths.

**Verify status: unavailable.** `factory-verify` rejected the repo (`"repo not allowed (rc/* only)"`), and direct `./mvnw` invocations were blocked by the permission layer in this session, so I could not run `mvn -pl mapsforge-poi test` here. I did a careful manual trace of the comparator and helper refactor instead — logic and imports check out, but the maintainer/CI should confirm the build actually goes green.

Source issue: cpesch/RouteConverter#312

---
**Builder stats** · model `sonnet` · confidence `0` · 4m 52s across 1 round(s) · 2 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/21346)

Closes #312

### DIFF
--- a/mapsforge-poi/src/main/java/slash/navigation/pois/mapsforge/MapsforgePoiLookup.java
+++ b/mapsforge-poi/src/main/java/slash/navigation/pois/mapsforge/MapsforgePoiLookup.java
@@ -218,8 +218,8 @@ class MapsforgePoiLookup {
 
     private BoundsAndVersion readBoundsAndVersion(File file) {
         try (Connection connection = open(file)) {
-            int specVersion = readVersion(connection);
             BoundingBox fromMetadata = readBoundsFromMetadata(connection);
+            int specVersion = readVersion(connection);
             BoundingBox bounds = fromMetadata != null ? fromMetadata : readBoundsFromIndex(connection, specVersion);
             return new BoundsAndVersion(bounds, specVersion);
         } catch (SQLException e) {
@@ -261,8 +261,9 @@ class MapsforgePoiLookup {
     private int readVersion(Connection connection) {
         try (Statement statement = connection.createStatement();
              ResultSet resultSet = statement.executeQuery("SELECT value FROM metadata WHERE name = 'version'")) {
-            if (resultSet.next())
-                return Integer.parseInt(resultSet.getString("value").trim());
+            String value = resultSet.next() ? resultSet.getString("value") : null;
+            if (value != null)
+                return Integer.parseInt(value.trim());
         } catch (SQLException | NumberFormatException e) {
             log.log(Level.FINE, "Cannot read POI database version", e);
         }
--- a/mapsforge-poi/src/main/java/slash/navigation/pois/mapsforge/MapsforgePoiLookup.java
+++ b/mapsforge-poi/src/main/java/slash/navigation/pois/mapsforge/MapsforgePoiLookup.java
@@ -117,6 +117,10 @@ class MapsforgePoiLookup {
             "GROUP BY poi_index.id, poi_data.data " +
             "LIMIT ?";
 
+    // Datasource ids known to serve the v4 schema; every id not listed here defaults to version 3.
+    private static final Map<String, Integer> DATASOURCE_SPEC_VERSIONS = Map.of("mapsforge-pois-4", 4);
+    private static final int DEFAULT_SPEC_VERSION = 3;
+
     private final DataSourceManager dataSourceManager;
 
     MapsforgePoiLookup(DataSourceManager dataSourceManager) {
@@ -146,23 +150,34 @@ class MapsforgePoiLookup {
         for (DataSource dataSource : dataSourceManager.getDataSourceService().getDataSources()) {
             for (slash.navigation.datasources.File file : dataSource.getFiles()) {
                 if (DOT_POI.equals(getExtension(file.getUri())) && matches(file.getBoundingBox(), bounds))
-                    descriptors.add(new PoiDescriptor(null, file, file.getBoundingBox(), dataSource.getName()));
+                    descriptors.add(new PoiDescriptor(null, file, file.getBoundingBox(),
+                            DATASOURCE_SPEC_VERSIONS.getOrDefault(dataSource.getId(), DEFAULT_SPEC_VERSION), dataSource.getName()));
             }
         }
-        descriptors.sort((d1, d2) -> {
-            if (d1.localFile() != null && d2.localFile() == null)
-                return -1;
-            if (d1.localFile() == null && d2.localFile() != null)
-                return 1;
+        descriptors.sort(descriptorPreference());
+        return descriptors;
+    }
+
+    static Comparator<PoiDescriptor> descriptorPreference() {
+        return (d1, d2) -> {
             if (d1.boundingBox() == null && d2.boundingBox() != null)
                 return 1;
             if (d1.boundingBox() != null && d2.boundingBox() == null)
                 return -1;
-            if (d1.boundingBox() == null)
-                return 0;
-            return Double.compare(d1.boundingBox().getSquareSize(), d2.boundingBox().getSquareSize());
-        });
-        return descriptors;
+            if (d1.boundingBox() != null) {
+                int bySize = Double.compare(d1.boundingBox().getSquareSize(), d2.boundingBox().getSquareSize());
+                if (bySize != 0)
+                    return bySize;
+            }
+            int byVersion = Integer.compare(d2.specVersion(), d1.specVersion());
+            if (byVersion != 0)
+                return byVersion;
+            if (d1.localFile() != null && d2.localFile() == null)
+                return -1;
+            if (d1.localFile() == null && d2.localFile() != null)
+                return 1;
+            return 0;
+        };
     }
 
     private List<PoiDescriptor> collectLocalPoiDescriptors(BoundingBox bounds) {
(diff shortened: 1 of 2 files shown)


cpesch:

correctness: Two runtime-behavior risks: the remote-version lookup keys off dataSource.getId() while the same call site's dataSource.getName() (and the new tests) suggest the id/name distinction is untested and could silently defeat the fix; and readVersion(connection) is now called unconditionally before the metadata-bounds read, changing the failure mode for local files whose version query fails but whose metadata bounds would otherwise resolve.

### mapsforge-poi/src/main/java/slash/navigation/pois/mapsforge/MapsforgePoiLookup.java

- **major** [correctness] line 154 — DATASOURCE_SPEC_VERSIONS is keyed by dataSource.getId(), but the sibling call on the same line uses dataSource.getName() for the descriptor's dataSourceName, and every new unit test bypasses this lookup entirely (they construct PoiDescriptor directly with a hardcoded specVersion int, e.g. 'mapsforge-pois-4' as a name-like literal, never through a real DataSource whose getId() is exercised). If DataSource.getId() does not return the literal 'mapsforge-pois-4' (e.g. that string is actually the name, not the id), every remote v4 lookup silently falls back to DEFAULT_SPEC_VERSION=3, and the shadowing bug from issue #312 is not actually fixed — with nothing in this diff (tests or otherwise) that would catch it.  _(correctness)_

```suggestion
Verify against the DataSource class what getId() actually returns for the mapsforge-pois-4 source, or key the map by getName() to match the existing dataSourceName usage; add a test that goes through the real DataSource object (or a stub with a getId()) rather than constructing PoiDescriptor directly.
```
- **major** [correctness] line 221 — readVersion(connection) is now invoked unconditionally at the top of readBoundsAndVersion, before readBoundsFromMetadata runs. Previously the version query only ran lazily inside readBoundsFromIndex, and only when metadata bounds were null — so a SQLException from readVersion could never suppress an otherwise-successful metadata-bounds read. Now, if readVersion throws for a local .poi file (e.g. one missing whatever version marker it queries) whose metadata bounds would still parse fine, the whole method falls into the catch block and returns null bounds instead of the previously-valid metadata bounds — a regression for such files, since a null bounds descriptor is dropped by matches().  _(correctness)_

```suggestion
If readVersion is known to never throw for any file that readBoundsFromMetadata can successfully parse, note that invariant in a comment; otherwise read metadata bounds first and only let a version-read failure fall back to DEFAULT_SPEC_VERSION without discarding already-successful bounds.
```

---

cpesch:
🤖 Patcher pushed a fix to `agent/builder/312-rc-site-28-prefer-the-newest-poi-spec-ve`.

cpesch:

correctness: The comparator refactor (bbox size -> specVersion descending -> local-before-remote) correctly fixes the shadowing bug described in #312, and readVersion/readBoundsAndVersion null-handling is sound. Main risk is that this PR explicitly could not run Maven, and it adds symbol references (a new static import and two bare generic types) that are unverified against the real classpath — this repo has shipped an unverified compile break from a builder PR before (#43/#48, slash.common.io.CompactCalendar).

### mapsforge-poi/src/main/java/slash/navigation/pois/mapsforge/MapsforgePoiLookup.java

- **minor** [correctness] line 121 — `Map.of(...)` at this line and the new `Comparator<PoiDescriptor>` return type at line 161 both name JDK types by simple name for the first time in this class (the prior comparator was an unannotated lambda, so no explicit Comparator reference existed) — cannot confirm from the diff whether `java.util.Map`/`java.util.Comparator` are already imported, and no build ran to catch it if not.  _(correctness)_

```suggestion
Confirm both imports are present (or add them) and that the module compiles.
```

### mapsforge-poi/src/test/java/slash/navigation/pois/mapsforge/MapsforgePoiLookupTest.java

- **major** [correctness] line 30 — New static import `slash.common.io.Files.recursiveDelete` is unverified — the PR body states Maven verification was unavailable, and this exact failure class (a builder PR referencing a nonexistent slash.common.io.* member) previously broke the build undetected for a day (RouteConverter #43/#48).  _(correctness)_

```suggestion
Run `mvn -pl mapsforge-poi test-compile` (or equivalent) before merge to confirm `slash.common.io.Files#recursiveDelete(File)` actually exists with this signature.
```

---

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: Both findings turned out to be false positives on inspection, so I made no edits.

Both reviewer findings verified as non-issues; no changes needed

## Summary

- **Finding 1 (major):** `slash.common.io.Files.recursiveDelete(File)` exists exactly as used — `common/src/main/java/slash/common/io/Files.java:473`, `public static void recursiveDelete(File path) throws IOException`. The test's static import and its `@After` method (declared `throws Exception`, a valid supertype of `IOException`) compile cleanly against it. No fix needed.
- **Finding 2 (minor):** `MapsforgePoiLookup.java` already has a wildcard import `java.util.*` at line 35, which covers both `Map` (used in `Map.of(...)` at line 121) and `Comparator` (used as the `descriptorPreference()` return type at line 161). No missing import — no fix needed.

Java/Maven remain unavailable in this environment (confirmed again: `java` not on `PATH`, `~/.sdkman` outside the allowed working directory), so I could not run an actual compile, but both concerns were about specific, checkable facts (a method's existence/signature, and an existing wildcard import) that I verified directly by reading the source rather than by inference. Working tree is unchanged (`git status` clean).
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-21361` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-21361).

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: No edits made — I did a thorough manual audit and couldn't find a fixable root cause, and couldn't execute the build to confirm one either. Here's my summary:

No further fix identified after manual review; build execution blocked in this session

## Summary

I inspected both commits on this branch — the original feature commit (`459ce2285`) and the prior patcher fix (`4f7a721ac`, which corrected a real NPE hazard in `readVersion` when the `version` metadata value is SQL `NULL`) — and traced the entire changed surface by hand:

- `MapsforgePoiLookup.java`: `descriptorPreference()` comparator, `readBoundsAndVersion`/`readBoundsFromMetadata`/`readBoundsFromIndex`/`readVersion`, and the `DATASOURCE_SPEC_VERSIONS` lookup.
- `MapsforgePoiLookupTest.java`: all 6 comparator-only tests plus the new end-to-end test `remoteDataSourceKnownToServeV4IsPreferredOverLocalV3ViaRealDataSourceId` (mock `DataSource`/`DataSourceManager` wiring, real SQLite file creation, `PoiDescriptor` field order, `BoundingBox` constructor/`getSquareSize()`/`intersect()` semantics, `Directories.getApplicationDirectory` directory creation, `Files.collectFiles`/`recursiveDelete`).

Every path traces correctly against the test assertions; I could not find a second defect beyond the one already fixed in `4f7a721ac`.

**Blocker:** this session's Bash tool denies every attempt to actually execute code — `./mvnw -v`, `javac -version`, even a bare `env` all returned "This command requires approval" with no user available to grant it (matching what the PR body and prior patch already reported: Maven is genuinely unavailable at the permission layer here). I therefore cannot reproduce the GitHub Actions failure (`log_tail` was also empty, so I have no excerpt to work from) or verify any change I might make.

Per the task's own guidance ("If no safe fix is possible, make NO edits and explain why"), I'm leaving the working tree untouched rather than guessing at a change to code that already checks out logically c
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-21362` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-21362).

PR #320, opened 2026-08-22
Fix the ClassFormatError from double GUI Designer instrumentation in IDE builds

cpesch:
Local IntelliJ builds crash on startup with:

```
java.lang.ClassFormatError: Duplicate method name "$$$getFont$$$" with signature
"(Ljava.lang.String;IILjava.awt.Font;)Ljava.awt.Font;" in class file
slash/navigation/mapview/mapsforge/MapSelector
```

thrown from `MapsforgeMapView.initializeMapView(MapsforgeMapView.java:418)`.

### Cause

The form-backed classes keep their generated setup code in the sources — `MapSelector.java` defines `$$$setupUI$$$`, `$$$getFont$$$`, `$$$loadLabelText$$$` and friends, committed since `ab85baab9`. The build has no `javac2` / `ideauidesigner-maven-plugin`, only the `forms-rt` runtime dependency, so Maven compiles those methods once and never instruments. That is why CI and the nightly are fine.

`.idea/` is gitignored and carried no `uidesigner-configuration` component, so IntelliJ used its default — *Generate GUI into: binary class files* — and ran an instrumentation pass over classes that already contained the generated code. `javap -p` on the resulting `mapsforge-mapview/target/classes/.../MapSelector.class` shows two `$$$getFont$$$` methods, hence the `ClassFormatError`.

### Fix

Track `.idea/uiDesigner.xml` with the setting pinned, so a fresh checkout does not depend on the IDE default:

```xml
<component name="uidesigner-configuration">
  <option name="INSTRUMENT_CLASSES" value="false" />
</component>
```

The file is force-added; the rest of `.idea/` stays ignored (a `!.idea/uiDesigner.xml` negation would not work under the `.idea/` directory ignore anyway).

### After merging

Existing working copies still hold the doubly-instrumented classes. Reload the project so the new setting is picked up, then `mvn clean` (or Build → Rebuild Project). Verify with:

```
javap -p mapsforge-mapview/target/classes/slash/navigation/mapview/mapsforge/MapSelector.class | grep -c 'getFont'
```

which should print `1`.


cpesch (commented):

correctness: Diff only adds .idea/uiDesigner.xml (IDE project config plus IntelliJ's standard auto-generated Swing palette block); the INSTRUMENT_CLASSES=false option is the correct pin for the described root cause (source-generated $$$setupUI$$$ methods should not also be binary-instrumented) and there is no executable code path for a runtime bug.

---

cpesch (commented):

correctness: Configuration-only fix to prevent IntelliJ GUI Designer double-instrumentation. Root cause is correctly diagnosed: form classes have generated methods in source, Maven compiles once without instrumentation, but IntelliJ's default was to instrument binaries afterward, creating duplicates. Setting INSTRUMENT_CLASSES=false prevents the instrumentation pass. No code changes, no runtime logic, no contract breaks.

---

PR #322, opened 2026-08-22
"?" button opens /help/ instead of the dialog's own topic

### LINKED ISSUE #321: [help] "?" button opens /help/ instead of the dialog's own topic
## Refactor: "?" button and F1 must resolve the same help topic id

**Source:** maintainer report 2026-08-22 (verbatim): "the ? buttons and F1s in RouteConverter all point to /help". Goal: a dialog's `?` button opens that dialog's own help topic instead of the table of contents.
This spec is fully resolved (grilled 2026-08-22); implement without further questions.

### Root cause (measured)

`HelpManager.helpButton(JComponent owner)` reads `owner.getName()`. All 18 call sites pass `contentPane`, whose name is null — so every `?` click falls through to `openContents()` → `https://www.routeconverter.com/help/`. The topic id each dialog already carries is set on the `JDialog` itself (`SimpleDialog(Window, String name)`), two levels above the content pane, and only F1's `openTopicForComponent` walk ever finds it.

Second, smaller hole in the same method: `helpButton` calls `button.setName("help-button")`. Since `Component.getName()` **is** the topic-id carrier (specs/00030 §11.2), pressing F1 while the `?` button has focus resolves the topic `help-button` → `/help/help-button/` → 404. Nothing reads that name (grepped the whole tree: the only other occurrence is a quote in `rc-meta/specs/done/00074`).

### Scope

- Extract the ancestor walk into one reusable resolver; both `helpButton` and the F1 listener use it.
- Stop naming the `?` button.
- Delete the dead JavaHelp search index left behind by the phase-B `javax.help` removal.

Out of scope: explicit per-dialog topic ids on the content pane (decided against — the dialog name is already the id); per-tab topics inside `OptionsDialog` (own follow-up issue; `OptionsDialog` has previously overflowed the builder seed cap); a tooltip or mnemonic on the `?` button; the missing help pages themselves and the server-side unknown-topic fallback (separate `rc-content` / `rc-meta` issues).

### Files

- `common-gui/src/main/java/slash/navigation/gui/HelpManager.java`
- `common-gui/src/test/java/slash/navigation/gui/HelpManagerTest.java` (new)
- delete the directory `route-converter-gui/src/main/resources/slash/navigation/converter/gui/help/` (contains only `JavaHelpSearch/`; `JavaHelpSearch`, `javahelp` and `.hs` have zero references in the tree)

1. Add `static String resolveTopicId(Component start)` to `HelpManager`: walk `start` and its `getParent()` chain; skip any `JLayeredPane` in the chain (the existing `null.layeredPane` guard from #303); return the first non-null, non-empty `getName()`; return `null` when the chain is exhausted. Keep it `static` and free of instance state so it is unit-testable.
2. `openTopicForComponent(Component)` becomes: resolve via `resolveTopicId`; non-null → `openTopic(topic)`, null → `openContents()`.
3. `helpButton(JComponent owner)`: the action listener calls `openTopicForComponent(owner)`. Do **not** read `owner.getName()` directly.
4. Remove the `button.setName("help-button")` line. Do not replace it with another `setName` — if a test handle is ever needed, use `putClientProperty`.
5. No change to `openTopic` / `openContents` / `resolveBaseUrl` / `installF1KeyListener` URL construction, and no change to any dialog class.

Topic ids this lights up, all already set by `SimpleDialog` — `about`, `add-file`, `add-url`, `complete-flightplan`, `delete-positions`, `departure-time`, `downloads`, `find-place`, `insert-positions`, `login`, `maximum-position-count`, `options`, `rename`, `send-error-report` (route-converter-gui) plus `maps`, `themes`, `style` (route-converter).

**Do not rename any of those ids to prettier slugs.** `SimpleDialog.showWithPreferences()` uses `getName()` as the window-bounds preference prefix, so a rename silently discards every user's stored dialog position and size. The two awkward ids (`style`, `complete-flightplan`) are handled by aliases on the content side.

`HelpManager.resolveTopicId(Component)` — pure, needs no display. `JPanel`, `JLayeredPane`, `JRootPane` and `java.awt.Container` all construct headless; the existing `JMenuHelperTest` proves Swing component construction runs in this repo's CI.

`common-gui/src/test/java/slash/navigation/gui/HelpManagerTest.java` must cover:

- a named component returns its own name
- an unnamed child returns the nearest named ancestor
- a **named** `JLayeredPane` in the chain is skipped and the walk continues past it
- an empty-string name is treated as absent
- a fully unnamed chain returns `null`
- the reported-bug shape: `Container` named `"options"` → `JRootPane` → `rootPane.getContentPane()` (unnamed) → `JButton` (unnamed) resolves to `"options"`

### Acceptance criteria

- `mvn -pl common-gui test` green including the new `HelpManagerTest`; full `mvn -q -DskipTests package` green.
- Manual: start with `-Dhelp.dryRun=true`, open **Extras → Options…**, click `?` → console prints `HelpManager dry-run: https://www.routeconverter.com/help/options/` (not `.../help/`).
- Manual: focus the `?` button in any dialog, press F1 → same URL, not `/help/help-button/`.

cpesch:
All changes are in place. Final summary:

## Summary

Fixed `HelpManager` so the `?` button opens each dialog's own help topic instead of always falling through to the table of contents, matching what F1 already did.

**Root cause:** `helpButton` read `owner.getName()` directly, but callers pass `contentPane`, which is unnamed — the topic id lives two levels up on the `JDialog` itself. Only F1's ancestor-walking listener ever found it.

**Changes (`common-gui/src/main/java/slash/navigation/gui/HelpManager.java`):**
- Extracted the ancestor walk (skip `JLayeredPane`, return first non-null/non-empty `getName()`, else `null`) into a new package-private static `resolveTopicId(Component)` — pure and unit-testable.
- `openTopicForComponent` now delegates to `resolveTopicId`.
- `helpButton`'s action listener now calls `openTopicForComponent(owner)` instead of reading `owner.getName()` directly, so `?` and F1 resolve identically.
- Removed `button.setName("help-button")` — that string was being picked up by F1's resolver as a bogus topic id, sending it to `/help/help-button/` (404) whenever the `?` button had focus. Not replaced with any other `setName`.

**New test:** `common-gui/src/test/java/slash/navigation/gui/HelpManagerTest.java` — covers all cases from the spec: own name, nearest named ancestor, named `JLayeredPane` skipped, empty-string name treated as absent, fully unnamed chain → `null`, and the reported bug shape (`Container("options")` → `JRootPane` → unnamed content pane → unnamed `JButton` → resolves to `"options"`).

**Deleted:** `route-converter-gui/src/main/resources/slash/navigation/converter/gui/help/` (only contained `JavaHelpSearch/`, confirmed zero references anywhere in the tree).

**Verification:** `factory-verify` is unavailable for this repo (broker rejects non-`rc/*` repos, exit code 2) and other `mvnw` invocations required approval I couldn't obtain in this session, so the build/tests were **not run** — I verified the change by careful code reading only (confirmed no other file references `"help-button"`, and traced the `JRootPane`/`getContentPane()`/`JLayeredPane` hierarchy used in the new test against actual Swing internals). Please run `mvn -pl common-gui test` before merging.

Source issue: cpesch/RouteConverter#321

---
**Builder stats** · model `sonnet` · confidence `0` · 1m 39s across 1 round(s) · 4 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/22455)

Closes #321


common-gui/src/main/java/slash/navigation/gui/HelpManager.java
```diff
@@ -65,11 +65,7 @@ private static String stripTrailingSlash(String url) {
 
     public JButton helpButton(final JComponent owner) {
         JButton button = new JButton("?");
```
cpesch, line 67: **nit** [correctness] Removing button.setName("help-button") is correct for the reported bug (it was the bogus id F1's resolver picked up), but this diff slice can't confirm nothing else locates this button by name (e.g. a Swing UI test using findComponentByName). Low confidence since such a reference, if it exists, is outside the files included here.

cpesch (commented):

correctness: The ancestor-walk extraction is a correct, behavior-preserving refactor for F1 and correctly fixes the reported '?' button bug (it now walks owner's ancestors instead of reading owner.getName() directly). Two lower-confidence risks outside the core fix are noted below.

### common-gui/src/main/java/slash/navigation/gui/HelpManager.java

- **nit** [correctness] line 67 — Removing button.setName("help-button") is correct for the reported bug (it was the bogus id F1's resolver picked up), but this diff slice can't confirm nothing else locates this button by name (e.g. a Swing UI test using findComponentByName). Low confidence since such a reference, if it exists, is outside the files included here.  _(correctness)_

### route-converter-gui/src/main/resources/slash/navigation/converter/gui/help/JavaHelpSearch/POSITIONS

- **minor** [correctness] — Only POSITIONS and TMAP are deleted from the JavaHelpSearch directory; a JavaHelp full-text search index is normally several coupled files (e.g. DOCS, DOCS.TAB, OFFSETS, SCHEMA, STOPWORDS, TMAP, POSITIONS). If any sibling index files still exist in the tree (not visible in this diff/files_changed) and a helpset.hs still declares a search view over this merge directory, this leaves a partially-deleted index that could throw at runtime when a user opens Help search, rather than cleanly removing the feature.  _(correctness)_

```suggestion
Confirm via `git ls-tree` (not just this diff) that no other files remain under .../help/JavaHelpSearch/, and that no *.hs helpset declares a KeywordSearch/JavaHelpSearch view pointing at this merge directory.
```

---

cpesch:

**Reason:** agentic run made no working-tree change. Model summary: No safe fix identified — no edits made

I reviewed the full diff introduced in `f5dd24707` (`HelpManager.java` + `HelpManagerTest.java` + the deleted `JavaHelpSearch` resources) against the CI failure for PR 322, but the provided `log_tail` was empty and I have no network access to fetch the actual GitHub Actions log for run `32562401731`.

What I checked instead, by reading the code:

- `resolveTopicId`'s ancestor walk is logically correct: I traced the real Swing `JRootPane` hierarchy (`contentPane` → `layeredPane` (a `JLayeredPane`, skipped) → `JRootPane` → the enclosing `Container`/`JDialog`), and confirmed `JComponent`/`Container` don't auto-generate a `getName()` value (only AWT heavyweight widgets like `Window`/`Dialog`/`Button` do), so the new unit tests' assumptions hold.
- All three call sites (`RenameDialog`, `OptionsDialog`, `MapsDialog`, `ThemeStyleDialog`, `ThemesDialog`) pass a `JComponent` to `helpButton`, matching the unchanged signature — no call-site breakage.
- `SimpleDialog` already calls `setName(name)` on the dialog itself, which is exactly the ancestor the new resolver is designed to find — consistent with the PR's stated root cause.
- The deleted `route-converter-gui/.../help/JavaHelpSearch/*` resources have zero remaining references anywhere (code, pom.xml) — safe removal, no dangling build config.
- No new i18n keys, no raw types/unchecked/deprecated APIs, no unused imports, no duplicate classes — nothing that should trip `failOnWarning` or the CodeQL/i18n checks called out in `AGENTS.md`.

I could not execute `./mvnw` to reproduce the failure directly — the tool policy in this sandbox blocks any command touching outside `/tmp/target` (including `which java`, `ls /usr/lib/jvm`), so there's no JDK/Maven reachable here, matching what the PR author already reported in the same session. Without an actual compiler/test error to work from, I'd be guessing at a fix rather than addressing a root cause, so per the task instructions I'm making no ed
**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-22460` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-22460).

common-gui/src/main/java/slash/navigation/gui/HelpManager.java
```diff
@@ -65,11 +65,7 @@ private static String stripTrailingSlash(String url) {
 
     public JButton helpButton(final JComponent owner) {
```
cpesch, line 66: **nit** [correctness] button.setName("help-button") is removed with nothing substituted; if any other file (not visible in this diff) looks up this button by name (UI test harness, accessibility script), that lookup silently breaks — worth a quick repo-wide grep for "help-button" before merging.

cpesch (commented):

correctness: resolveTopicId's synthetic-name skip correctly neutralizes JRootPane's "null.contentPane"/"null.glassPane" auto-naming (verified against JRootPane's actual pane-creation behavior and all six new test cases); no runtime bugs found in the changed logic itself.

### common-gui/src/main/java/slash/navigation/gui/HelpManager.java

- **nit** [correctness] line 66 — button.setName("help-button") is removed with nothing substituted; if any other file (not visible in this diff) looks up this button by name (UI test harness, accessibility script), that lookup silently breaks — worth a quick repo-wide grep for "help-button" before merging.  _(correctness)_

---

PR #323, opened 2026-08-22
fix: an SVG icon that fails to load no longer breaks the map view for the whole session

cpesch:
## Problem

Observed while switching a position list between waypoints and tracks: every click on the map threw

```
java.lang.NullPointerException: Cannot invoke "org.mapsforge.core.graphics.Bitmap.getWidth()" because "this.bitmap" is null
	at org.mapsforge.map.layer.overlay.Marker.contains(Marker.java:58)
	at slash.navigation.mapview.mapsforge.overlays.DraggableMarker.onTap(DraggableMarker.java:48)
	at slash.navigation.mapview.mapsforge.helpers.MapViewMoverAndZoomer.getMarkerFor(MapViewMoverAndZoomer.java:127)
```

and the position list was no longer updated:

```
SCHWERWIEGEND: Cannot replace route: java.lang.NullPointerException
SCHWERWIEGEND: Cannot rebuild after java.lang.NullPointerException: java.lang.NullPointerException
WARNUNG: Could not find layer to remove for PositionWithLayer@...[position=..., layer=null]
WARNUNG: Cannot remove layers []
```

## Root cause

Two minutes earlier the log has the trigger:

```
SCHWERWIEGEND: Cannot create magnifier icon: java.io.IOException: java.io.IOException: Stream closed
SCHWERWIEGEND: Cannot create marker icon: java.io.IOException: java.io.IOException: Stream closed
```

* `markerIcon` was created once in an instance initializer, so after that single failure it stayed `null` for the whole session. `Marker` accepts a `null` bitmap, `Marker#draw` checks for it, but `Marker#contains` does not - so every `mousePressed` near a selected position crashed.
* `waypoint.svg` failed the same way. `SVGUniverse#loadSVG` returns `null` on failure, so `SVGHelper#getResourceBitmap` dereferenced a `null` `SVGDiagram`. That `NullPointerException` escaped `WaypointOperation#add` before `positionWithLayer.setLayer(marker)` ran, and `UpdateDecoupler#rebuild` failed the same way - hence the permanent `layer=null` warnings.
* A failed load also leaves a half loaded document in the `SVGUniverse` (only a `SAXParseException` removes it), so every later load of the same name fails too - the reason a single glitch never recovered.
* `UpdateDecoupler` logged only `e.toString()`, so the second `NullPointerException` had no stack trace at all, and the `com.kitfox` warning naming the failing SVG was filtered out of the log file.

## Changes

* `DraggableMarker#onTap` - no hit test without a bitmap
* `SVGHelper` - throw an `IOException` instead of dereferencing null, and drop the half loaded document so a later load retries
* `MapsforgeMapView` - marker, magnifier and waypoint icons are created on demand and retried after a failure; a failed `renderSvg` purges its cache entry
* `UpdateDecoupler` - log the stack trace
* `LoggingHelper` - pass `com.kitfox` records into the log file

## Verification

`javac` over `mapsforge-mapview` and `mvn compile` for `common` are clean (the reactor build of `mapsforge-mapview` fails only on unrelated uncommitted `MapSelector.java` warnings in my working tree). Behaviour verified by reading the failure path; the original glitch (an unreadable classpath resource stream, most likely the module rebuilt while the app was running) is not reproducible on demand.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
@@ -123,6 +123,7 @@ import static slash.navigation.mapview.mapsforge.helpers.MapViewCalculations.col
 import static slash.navigation.mapview.mapsforge.helpers.MapViewCalculations.computeAddRow;
 import static slash.navigation.mapview.mapsforge.helpers.MapViewCalculations.thresholdForPixel;
 import static slash.navigation.mapview.mapsforge.helpers.SVGHelper.getResourceBitmap;
+import static slash.navigation.mapview.mapsforge.helpers.SVGHelper.removeDocument;
 import static slash.navigation.mapview.mapsforge.helpers.WithLayerHelper.*;
 import static slash.navigation.mapview.mapsforge.models.LocalNames.MAP;
 
@@ -229,16 +230,22 @@ public class MapsforgeMapView extends BaseMapView {
 
         this.selectionUpdater = new SelectionUpdater(positionsModel, new SelectionOperation() {
             private Bitmap markerIcon;
-            {
-                try {
-                    markerIcon = GRAPHIC_FACTORY.renderSvg(MapsforgeMapView.class.getResourceAsStream("marker.svg"), 1.0f, 32, 54, 100, 1234567892);
-                } catch (IOException e) {
-                    log.severe("Cannot create marker icon: " + e);
+
+            // created on demand and retried after a failure: an icon that failed to load once
+            // used to leave the session with markers that cannot be drawn or tapped at all
+            private synchronized Bitmap getMarkerIcon() {
+                if (markerIcon == null) {
+                    try {
+                        markerIcon = createIcon("marker.svg", MARKER_ICON_HASH, 32, 54);
+                    } catch (IOException e) {
+                        log.severe("Cannot create marker icon: " + e);
+                    }
                 }
+                return markerIcon;
             }
 
             private Marker createMarker(PositionWithLayer positionWithLayer, LatLong latLong) {
-                return new DraggableMarker(MapsforgeMapView.this, positionWithLayer, latLong, markerIcon, 0, -25);
+                return new DraggableMarker(MapsforgeMapView.this, positionWithLayer, latLong, getMarkerIcon(), 0, -25);
             }
 
             public void add(List<PositionWithLayer> positionWithLayers) {
@@ -540,6 +547,19 @@ public class MapsforgeMapView extends BaseMapView {
         return mapView;
     }
 
+    private static final int MAGNIFIER_ICON_HASH = 1234567891;
+    private static final int MARKER_ICON_HASH = 1234567892;
+
+    private static Bitmap createIcon(String resource, int hash, int width, int height) throws IOException {
+        try {
+            return GRAPHIC_FACTORY.renderSvg(MapsforgeMapView.class.getResourceAsStream(resource), 1.0f, width, height, 100, hash);
+        } catch (IOException e) {
+            // a failed load may leave a half loaded document behind that fails every later load, too
+            removeDocument(Integer.toString(hash));
+            throw e;
+        }
+    }
+
     private Bitmap waypointIcon;
 
     private synchronized Bitmap createWaypointIcon() {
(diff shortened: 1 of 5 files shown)


mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
```diff
@@ -540,6 +547,19 @@ private AwtGraphicMapView createMapView() {
...
 
+    private static final int MAGNIFIER_ICON_HASH = 1234567891;
+    private static final int MARKER_ICON_HASH = 1234567892;
+
+    private static Bitmap createIcon(String resource, int hash, int width, int height) throws IOException {
+        try {
+            return GRAPHIC_FACTORY.renderSvg(MapsforgeMapView.class.getResourceAsStream(resource), 1.0f, width, height, 100, hash);
+        } catch (IOException e) {
+            // a failed load may leave a half loaded document behind that fails every later load, too
+            removeDocument(Integer.toString(hash));
```
cpesch, line 558: **minor** [correctness] removeDocument(Integer.toString(hash)) purges a document from SVGCache.getSVGUniverse(), but createIcon()'s failing call is GRAPHIC_FACTORY.renderSvg(...) — a mapsforge-library method whose internal SVG cache/universe instance is not shown in this diff. If it doesn't use the same shared universe, this purge is a no-op against the actual poisoned document and the marker/magnifier icons will keep failing to load on every retry, unlike the waypoint path (SVGHelper) where the same universe is unambiguously the one loadSVG just used.
```suggestion
Confirm (e.g. via mapsforge-map-awt source or a test that re-renders after a forced parse failure) that GRAPHIC_FACTORY.renderSvg and SVGCache.getSVGUniverse() share the same SVGUniverse instance; if not, purge whatever cache renderSvg actually uses, or drop the misleading removeDocument call from createIcon().
```

mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
```diff
@@ -229,16 +230,22 @@ public void initialize(PositionsModel positionsModel,
...
             private Bitmap markerIcon;
-            {
-                try {
-                    markerIcon = GRAPHIC_FACTORY.renderSvg(MapsforgeMapView.class.getResourceAsStream("marker.svg"), 1.0f, 32, 54, 100, 1234567892);
-                } catch (IOException e) {
-                    log.severe("Cannot create marker icon: " + e);
+
+            // created on demand and retried after a failure: an icon that failed to load once
+            // used to leave the session with markers that cannot be drawn or tapped at all
+            private synchronized Bitmap getMarkerIcon() {
```
cpesch, line 236: **minor** [correctness] getMarkerIcon()/getMagnifierIcon() retry createIcon() synchronously on every call while the icon is null, with no backoff or failure caching. If the underlying resource load keeps failing (e.g. the cache-purge issue above means it never recovers), every marker creation / magnifier show will re-attempt SVG parsing and hit the renderSvg exception path again, which — depending on which thread creates markers/shows the magnifier — could repeatedly do blocking I/O and exception handling on the UI thread instead of failing fast after the first attempt.
```suggestion
Consider a short cooldown or single-attempt-per-session fallback after N consecutive failures, or confirm callers are never on the EDT.
```

cpesch (commented):

correctness: The waypoint icon fix (SVGHelper) is solid and directly closes the loop described in the PR; the marker/magnifier icon fix (MapsforgeMapView#createIcon) relies on removeDocument() targeting the same cache GRAPHIC_FACTORY.renderSvg uses internally, which is unverifiable from this diff and, if false, would leave those two icons unable to recover exactly as before (though no longer crashing, thanks to the null checks).

- **minor** [correctness] line 236 — getMarkerIcon()/getMagnifierIcon() retry createIcon() synchronously on every call while the icon is null, with no backoff or failure caching. If the underlying resource load keeps failing (e.g. the cache-purge issue above means it never recovers), every marker creation / magnifier show will re-attempt SVG parsing and hit the renderSvg exception path again, which — depending on which thread creates markers/shows the magnifier — could repeatedly do blocking I/O and exception handling on the UI thread instead of failing fast after the first attempt.  _(correctness)_

```suggestion
Consider a short cooldown or single-attempt-per-session fallback after N consecutive failures, or confirm callers are never on the EDT.
```
- **minor** [correctness] line 558 — removeDocument(Integer.toString(hash)) purges a document from SVGCache.getSVGUniverse(), but createIcon()'s failing call is GRAPHIC_FACTORY.renderSvg(...) — a mapsforge-library method whose internal SVG cache/universe instance is not shown in this diff. If it doesn't use the same shared universe, this purge is a no-op against the actual poisoned document and the marker/magnifier icons will keep failing to load on every retry, unlike the waypoint path (SVGHelper) where the same universe is unambiguously the one loadSVG just used.  _(correctness)_

```suggestion
Confirm (e.g. via mapsforge-map-awt source or a test that re-renders after a forced parse failure) that GRAPHIC_FACTORY.renderSvg and SVGCache.getSVGUniverse() share the same SVGUniverse instance; if not, purge whatever cache renderSvg actually uses, or drop the misleading removeDocument call from createIcon().
```

---

PR #324, opened 2026-08-22
fix: wire the "?" button of every dialog, not just DownloadsDialog

cpesch:
## Problem

Every dialog wired its help button *inside* the GUI Designer generated `$$$setupUI$$$`:

```java
buttonHelp = BaseRouteConverter.getInstance().getContext().getHelpManager().helpButton(contentPane);
```

That method is regenerated from the `.form` whenever the form is touched in the IDE, and regeneration replaces the call with a plain `new JButton()` + `setText("?")` — no listener. DownloadsDialog is where it happened: the `?` button was dead, and F1 was the only way to reach the topic.

## Fix

`SimpleDialog.setContentPane` walks the content pane and hands every still-unwired `JButton` labelled `"?"` to the new `HelpManager.registerHelpButton(button, owner)`. Every dialog already calls `setContentPane(contentPane)` in its constructor, so all 18 forms are covered, the hook lives outside the generated method, and a new dialog needs nothing beyond the button in its form.

The follow-up commit drops the now-redundant `helpButton(contentPane)` call from all 17 remaining dialogs' generated methods, so no generated code carries wiring any more. `HelpManager.helpButton` goes away with its last call site; `registerHelpButton` replaces it.

The topic id still comes from the dialog's `getName()` via `HelpManager.resolveTopicId`, so each dialog opens its own page rather than the help contents.

DownloadsDialog additionally gets its buttons into a single row (`?` left, `Schließen` right) instead of the `?`/restart/stop/remove row plus a separate `Schließen` row.

## Verification

Ran the GUI with `-Dhelp.dryRun=true` and clicked each dialog's `?` from an injected agent:

```
downloads listeners=1  → https://www.routeconverter.de/help/downloads/
options   listeners=1  → https://www.routeconverter.de/help/options/
themes    listeners=1  → https://www.routeconverter.de/help/themes/
style     listeners=1  → https://www.routeconverter.de/help/style/
about     listeners=1  → https://www.routeconverter.de/help/about/
```

One button, one listener, one browser open per click — the `getActionListeners().length == 0` guard means a dialog that wires its own button is left alone. Downloads button row measured live: all five buttons at the same `y`.

### DIFF
--- a/common-gui/src/main/java/slash/navigation/gui/HelpManager.java
+++ b/common-gui/src/main/java/slash/navigation/gui/HelpManager.java
@@ -63,16 +63,17 @@ public class HelpManager {
     public void openTopic(String topicId) { browse(resolveBaseUrl() + "/help/" + topicId + "/"); }
     public void openContents() { browse(resolveBaseUrl() + "/help/"); }
 
-    public JButton helpButton(final JComponent owner) {
-        JButton button = new JButton("?");
+    /**
+     * Wires an existing {@code ?} button to the help topic of {@code owner}. A button declared in a
+     * GUI Designer form is instantiated by the generated {@code $$$setupUI$$$} method.
+     */
+    public void registerHelpButton(final AbstractButton button, final JComponent owner) {
         button.addActionListener(e -> openTopicForComponent(owner));
-        return button;
     }
 
     public void installF1KeyListener() {
         Toolkit.getDefaultToolkit().addAWTEventListener(event -> {
-            if (!(event instanceof java.awt.event.KeyEvent)) return;
-            java.awt.event.KeyEvent ke = (java.awt.event.KeyEvent) event;
+            if (!(event instanceof java.awt.event.KeyEvent ke)) return;
             if (ke.getID() != KEY_PRESSED || ke.getKeyCode() != VK_F1) return;
             openTopicForComponent(ke.getComponent());
             ke.consume();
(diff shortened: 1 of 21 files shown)


cpesch (commented):

correctness: Wiring change looks internally consistent: HelpManager.registerHelpButton() replaces the removed helpButton() factory method, its sole prior caller (DownloadsDialog) is updated in the same diff, and SimpleDialog.setContentPane()'s recursive scan runs after super.setContentPane() so the button is already in the hierarchy when owner-resolution happens (consistent with the existing F1-key path which already passes arbitrary descendant components, not the dialog itself, as the owner). DownloadsDialog.form/.java row/column indices for the merged button row match between the two files. No bugs found in the diff's own lines; the other 17 dialogs that also gain wiring are unchanged in this diff and unverifiable from here, but the PR body's live listener-count evidence for 4 of them is consistent with the mechanism working.

---

cpesch:
🤖 Patcher pushed a fix to `fix/help-button-wiring-in-simpledialog`.

cpesch (commented):

correctness: Mechanical replacement of a returned help-button factory with a recursive auto-wiring hook in SimpleDialog.setContentPane; the owner-parameter substitution (button passed as its own owner) and the DownloadsDialog row-merge layout both check out against the diff and the PR's own runtime verification, and the listener-count guard correctly prevents double-wiring the one dialog that previously wired itself by hand.

---

PR #346, opened 2026-08-30
Delete download scratch temp files on failure/removal, not only on success

### LINKED ISSUE #342: [rc-support#21] Delete download scratch temp files on failure/removal, not only on success
## Bug: failed/removed downloads leak scratch temp files forever

**Source:** [rc-support#21](https://git.routeconverter.com/rc/rc-support/issues/21) (mail intake). Reporter (paraphrased, PII scrubbed): after redirecting RouteConverter's temp directory via Windows Registry, hundreds of leftover `*.xml` temp files accumulate there and are never cleaned up. Goal: stop orphaning the per-download scratch temp file on any failure/removal path, without touching legitimate resumable-partial-download behavior.

### Scope
`Download.newTempFile()` creates one scratch temp file per queued download (every `DataSourceManager`/`Edition` catalog XML re-fetch included). `GetPerformer.postProcess()` currently only deletes it on the full success path. Fix the three specific points confirmed safe without breaking resume:
1. `GetPerformer.postProcess()` — delete the temp file when `validate()` returns `false` (this point is only reached after a *fully completed* HTTP transfer — see Behaviour below — so a validation failure here means corrupt/wrong content, never an incomplete-and-resumable file).
2. `GetPerformer.run()` — delete the temp file (if it exists) in the final `else` branch, i.e. when both `resume()` and the `download()` fallback failed to produce a successful result this round.
3. `DownloadManager.removeDownloads()` — delete each download's temp file (if it exists) when the user explicitly removes it from the queue.

Out of scope (do NOT do these — they would break resume-across-restart):
- `DownloadManager.stopDownloads()` must NOT delete temp files — a stopped-but-not-removed download stays in the queue with state `Stopped` and is restartable via `restartDownloads()`, which resumes from the existing partial temp file.
- Any exception thrown out of `resume()`/`download()` mid-transfer (e.g. a network drop while `Copier.copyAndClose()` is writing) must NOT trigger cleanup — that partial file is exactly what a later resume attempt needs; do not wrap `run()`'s `resume()`/`download()` calls in a broad try/delete/rethrow.
- No periodic/startup sweep of the temp directory — that's a separate, larger feature (age-based reaper), deliberately deferred.

### Files
- `download/src/main/java/slash/navigation/download/performer/GetPerformer.java`
- `download/src/main/java/slash/navigation/download/DownloadManager.java`
- `download/src/test/java/slash/navigation/download/` (check for existing `GetPerformerTest`/`DownloadManagerTest`; none currently exist — add new test files in this package)

- **Why `postProcess()`'s `validate()==false` branch is safe to clean up unconditionally:** `run()` (`GetPerformer.java:133-157`) only calls `postProcess()` inside the `result.success` branch, and `result.success` is only ever `true` after `resume()`'s `isPartialContent()` fully copied the response body, or `download()`'s full-body copy completed without throwing. So by the time `validate()` runs, the transfer is complete — a checksum/size mismatch there is corruption of a whole file, never "needs another resume pass" (that's a separate signal, driven by `canResume()`'s content-length comparison, `GetPerformer.java:69-74`).
- `GetPerformer.postProcess()` (currently lines ~159-173): after `if (!validate()) return false;`, insert a temp-file delete before returning, e.g.:
  ```java
  if (!validate()) {
      deleteTempFileQuietly();
      return false;
  }
  ```
- `GetPerformer.run()` (currently lines ~133-157): in the trailing `else` branch (`downloadExecutor.downloadFailed();`), add the same cleanup call right after:
  ```java
  } else {
      downloadExecutor.downloadFailed();
      deleteTempFileQuietly();
  }
  ```
- Add a small private helper in `GetPerformer` used by both call sites:
  ```java
  private void deleteTempFileQuietly() {
      File tempFile = getDownload().getTempFile();
      if (tempFile.exists() && !tempFile.delete())
          log.warning(format("Cannot delete temp file %s", tempFile));
  }
  ```
  (Log-and-continue, not throw — matches the existing style of the rest of `GetPerformer`'s cleanup call inside `postProcess()`'s success path, but must not itself fail the download outcome that's already been decided.)
- `DownloadManager.removeDownloads()` (currently lines ~146-155): in the `for (Download download : downloads)` loop, after `stopDownloads(downloads)` and before/alongside `model.removeDownload(download)`, delete the temp file if present:
  ```java
  for (Download download : downloads) {
      log.info("Removing download " + download);
      File tempFile = download.getTempFile();
      if (tempFile.exists() && !tempFile.delete())
          log.warning("Cannot delete temp file " + tempFile);
      model.removeDownload(download);
  }
  ```
- Do not touch `stopDownloads()`, `restartDownloads()`, `canResume()`, or the `resume()`/`download()` HTTP methods.

### Testable seam
`download` module is a plain Java module (no GUI dependency) — no seam extraction needed, but there are currently no test files for `GetPerformer` or `DownloadManager`; add them.

JUnit cases:
- New `download/src/test/java/slash/navigation/download/performer/GetPerformerTest.java`: construct a `Download`/`GetPerformer` against a temp directory (use `@TempDir`), simulate a `validate()` failure (e.g. a `FileAndChecksum` with a checksum that won't match written bytes) and assert the temp file no longer exists after `postProcess()` returns `false`. If `GetPerformer`'s dependencies (`DownloadExecutor`, HTTP layer) make full instantiation impractical, at minimum unit-test the extracted `deleteTempFileQuietly()`-equivalent logic directly (widen visibility to package-private for the test if needed).
- New `download/src/test/java/slash/navigation/download/DownloadManagerTest.java`: `testRemoveDownloadsDeletesTempFile()` — create a `Download` whose `getTempFile()` points at a real temp file (create it via `@TempDir`), call `removeDownloads(singletonList(download))`, assert the file no longer exists.
- `testStopDownloadsDoesNotDeleteTempFile()` — same setup, call `stopDownloads(singletonList(download))` instead, assert the file still exists (regression guard for the explicit out-of-scope decision above).

### Acceptance criteria
- New tests pass; `mvn -pl download test` green.
- Manual check: let a datasource/edition catalog download fail validation (e.g. temporarily block network mid-refresh or point at a stale ETag causing a checksum mismatch) — its temp file is gone afterward, not left in the temp directory. Separately: stop (don't remove) an in-progress download, confirm its temp file is still present and the download resumes correctly on restart.

cpesch:
No hidden dependency on the temp file surviving removal — this is a simple UI action forwarding to `removeDownloads`. The implementation is complete and consistent with the rest of the codebase.

## Summary

Implemented the three fixes specified in the issue to stop orphaning per-download scratch temp files on failure/removal, without touching resume-across-restart behavior:

1. **`GetPerformer.postProcess()`** — deletes the temp file when `validate()` returns `false` (reached only after a fully completed transfer, per the issue's analysis).
2. **`GetPerformer.run()`** — deletes the temp file in the final `else` branch (both `resume()` and `download()` failed to produce a successful result).
3. **`DownloadManager.removeDownloads()`** — deletes each download's temp file when explicitly removed from the queue.

Added a shared private `deleteTempFileQuietly()` helper in `GetPerformer` (log-and-continue, matching existing style). Left `stopDownloads()`, `restartDownloads()`, `canResume()`, and the `resume()`/`download()` HTTP methods untouched, as required.

Added tests:
- `download/src/test/java/slash/navigation/download/performer/GetPerformerTest.java` — hermetic (local `HttpServer`) tests for a checksum-mismatch triggering `postProcess()` cleanup, and a failed transfer triggering `run()`'s cleanup of a stale pre-existing temp file.
- `download/src/test/java/slash/navigation/download/DownloadManagerTest.java` — `testRemoveDownloadsDeletesTempFile()` and the regression guard `testStopDownloadsDoesNotDeleteTempFile()`.

**Verify result: unavailable.** No JDK is present in this sandbox (`java`/`javac` not on `PATH`), and `factory-verify -pl download test` returned exit 2 with `repo not allowed (rc/* only)` — this repo isn't in the verify broker's allowed scope. I could not run `mvn -pl download test`. I instead did a careful manual trace of both new test files against the actual `GetPerformer`/`DownloadManager`/`Download` control flow (canResume/resume/download branching, model registration requirements for `updateDownload`/`removeDownload`, package-private access constraints) to confirm they exercise the intended paths and assertions correctly — but this is not a substitute for an actual test run, so please run `./mvnw -pl download test` to confirm before merging.

Source issue: cpesch/RouteConverter#342

---
**Builder stats** · model `sonnet` · confidence `0` · 6m 9s across 1 round(s) · 4 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/24812)

Closes #342

### DIFF
--- a/download/src/main/java/slash/navigation/download/DownloadManager.java
+++ b/download/src/main/java/slash/navigation/download/DownloadManager.java
@@ -148,6 +148,9 @@ public class DownloadManager {
 
         for (Download download : downloads) {
             log.info("Removing download " + download);
+            File tempFile = download.getTempFile();
+            if (tempFile.exists() && !tempFile.delete())
+                log.warning("Cannot delete temp file " + tempFile);
             model.removeDownload(download);
         }
 
--- a/download/src/main/java/slash/navigation/download/performer/GetPerformer.java
+++ b/download/src/main/java/slash/navigation/download/performer/GetPerformer.java
@@ -152,8 +152,10 @@ public class GetPerformer implements ActionPerformer {
             } else
                 downloadExecutor.postProcessFailed();
 
-        } else
+        } else {
             downloadExecutor.downloadFailed();
+            deleteTempFileQuietly();
+        }
     }
 
     private boolean postProcess(Long lastModified) throws IOException {
@@ -161,8 +163,10 @@ public class GetPerformer implements ActionPerformer {
 
         bringToTarget(lastModified);
 
-        if (!validate())
+        if (!validate()) {
+            deleteTempFileQuietly();
             return false;
+        }
 
         if (getDownload().getTempFile().exists())
             if (!getDownload().getTempFile().delete())
@@ -170,6 +174,12 @@ public class GetPerformer implements ActionPerformer {
 
         log.fine(format("Postprocess from %s successful", getDownload().getUrl()));
         return true;
+    }
+
+    private void deleteTempFileQuietly() {
+        File tempFile = getDownload().getTempFile();
+        if (tempFile.exists() && !tempFile.delete())
+            log.warning(format("Cannot delete temp file %s", tempFile));
     }
 
     private void bringToTarget(Long lastModified) throws IOException {
--- a/download/src/test/java/slash/navigation/download/DownloadManagerTest.java
+++ b/download/src/test/java/slash/navigation/download/DownloadManagerTest.java
@@ -1,0 +1,95 @@
+/*
+    This file is part of RouteConverter.
+
+    RouteConverter is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    RouteConverter is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with RouteConverter; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+    Copyright (C) 2007 Christian Pesch. All Rights Reserved.
+*/
+package slash.navigation.download;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+
+import static java.io.File.createTempFile;
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static slash.navigation.download.Action.Copy;
+import static slash.navigation.download.State.Queued;
+
+/**
+ * Tests that {@link DownloadManager} deletes a download's scratch temp file when it's explicitly
+ * removed from the queue, but never when it's merely stopped (a stopped-but-not-removed download
+ * stays restartable via {@link DownloadManager#restartDownloads} and resumes from that temp file).
+ *
+ * @author Christian Pesch
+ */
+public class DownloadManagerTest {
+    private File queueFile, target, tempFile;
+    private DownloadManager manager;
+
+    @Before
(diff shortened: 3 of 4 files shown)


cpesch (commented):

correctness: The implementation correctly deletes temp files on validation failure and download failure while preserving the existing success-path deletion. The cleanup in removeDownloads() is properly ordered before model removal. Tests cover the new failure paths and include a regression guard for stopDownloads(). No null-check or race-condition issues given Download's tempFile is always a valid File object and download queues are single-threaded per download.

---

PR #347, opened 2026-08-30
Fix UTF-8/ISO-8859-1 charset mismatch mangling non-ASCII descriptions on copy/paste

### LINKED ISSUE #340: [rc-support#21] Fix UTF-8/ISO-8859-1 charset mismatch mangling non-ASCII descriptions on copy/paste
## Bug: non-ASCII position descriptions get mangled (mojibake) on copy/paste between RouteConverter instances

**Source:** [rc-support#21](https://git.routeconverter.com/rc/rc-support/issues/21) (mail intake). Reporter (paraphrased, PII scrubbed): a position description containing "San Nicolò" survives copy to Notepad++ correctly, but pasting it back into RouteConverter (or copying between two RouteConverter instances) turns it into "San NicolÃ²" — happens on both German and Italian localized builds. Goal: fix the charset mismatch in the clipboard string round-trip so non-ASCII descriptions survive copy/paste losslessly.

### Scope
Fix the charset mismatch between how `PositionSelection`'s string-flavor clipboard payload is written (correct: char-based `PrintWriter`, no encoding issue) and how it's re-parsed on paste (`NavigationFormatParser.read(String)` → `source.getBytes()` using the JVM platform-default charset → decoded by `GlopusFormat`'s inherited `TextNavigationFormat.read(InputStream, ParserContext)`, hardcoded to `ISO_LATIN1_ENCODING`). The two sides use different, unmatched charsets — encode side is effectively UTF-8 (or worse, whatever Windows' platform-default charset happens to be), decode side is always ISO-8859-1.

Out of scope: touching any other `TextNavigationFormat` subclass's default encoding (other formats read real device/file exports where ISO-8859-1 is the correct legacy default — do not change the shared default) or the general `read(String)`/`read(InputStream)` API surface used by every other format.

### Files
- `navigation-formats/src/main/java/slash/navigation/base/NavigationFormatParser.java`
- `navigation-formats/src/main/java/slash/navigation/simple/GlopusFormat.java`
- `navigation-formats/src/test/java/slash/navigation/base/NavigationFormatParserTest.java`

- In `NavigationFormatParser.read(String source)` (currently `return read(new ByteArrayInputStream(source.getBytes()));`), change to explicitly encode as UTF-8: `source.getBytes(UTF8_ENCODING)` (constant already exists in `slash.common.io.Transfer`, import it). This removes the platform-default-charset dependency (which on some Windows locales is not UTF-8 at all, an additional latent bug beyond the reported one).
- `GlopusFormat` is RouteConverter's own internal clipboard/`.tk` format (see its class javadoc: "Reads and writes Glopus (.tk) files") — it doesn't inherit any external-file-compatibility constraint that requires ISO-8859-1. Override `read(InputStream, ParserContext)` in `GlopusFormat` to decode as UTF-8 instead of the inherited ISO-8859-1 default:
  ```java
  public void read(InputStream source, ParserContext<Wgs84Route> context) throws IOException {
      read(source, UTF8_ENCODING, context);
  }
  ```
  (mirrors the existing protected `read(InputStream, String encoding, ParserContext)` helper already in `TextNavigationFormat`; import `UTF8_ENCODING` from `slash.common.io.Transfer` alongside the existing static import).
- Do not add a general encoding-detection mechanism (e.g. BOM sniffing) — this is a two-sided, closed-loop fix: RouteConverter controls both the writer (`PositionSelection.createStringFor`, already charset-clean) and now the reader for this one format.
- Note: this changes how real on-disk `.tk` files are decoded too (was ISO-8859-1, now UTF-8) — acceptable, since the format's only in-repo producer/consumer is RouteConverter itself and the fix makes read/write symmetric; flag this explicitly in the PR description as an intentional side effect for the maintainer to note in the changelog.

### Testable seam
`navigation-formats` is a plain Java module, fully unit-testable headless — no GUI seam extraction needed.

JUnit cases:
- `NavigationFormatParserTest`: `testReadFromStringPreservesNonAsciiCharacters()` — call `new NavigationFormatParser().read("51.0450383,7.0508300,San Nicolò")` (Glopus comment-line format, see `GlopusFormat.COMMENT_LINE_PATTERN`), assert the parsed position's description equals `"San Nicolò"` exactly (round-trip through the fixed encode/decode pair).
- Add a `GlopusFormatTest` (or extend existing round-trip test if one exists — check `navigation-formats/src/test/java/slash/navigation/simple/` first) covering: write a route with a non-ASCII description via `GlopusFormat.write(...)` to a `StringWriter`, then read the resulting string back via `NavigationFormatParser.read(String)`, assert descriptions match.

### Acceptance criteria
- New tests pass; `mvn -pl navigation-formats test` green.
- Manual check: in RouteConverter, create a position with description "San Nicolò", copy it, paste it into a second RouteConverter instance (or the same instance after restart) — description reads "San Nicolò", not "San NicolÃ²".

cpesch:
All four files match the issue's file list. Here's the summary.

## Summary

Fixed the UTF-8/ISO-8859-1 charset mismatch that mangled non-ASCII position descriptions (e.g. "San Nicolò" → "San NicolÃ²") on copy/paste between RouteConverter instances.

**Root cause:** `PositionSelection`'s clipboard payload is written char-based (no encoding issue), but `NavigationFormatParser.read(String)` re-encoded it via `source.getBytes()` (platform-default charset) while `GlopusFormat` (inherited from `TextNavigationFormat`) always decoded as ISO-8859-1. The two sides never agreed on an encoding.

**Changes:**
- `NavigationFormatParser.java`: `read(String source)` now explicitly encodes with `UTF8_ENCODING` instead of the JVM platform-default charset.
- `GlopusFormat.java`: overrides `read(InputStream, ParserContext)` to decode with `UTF8_ENCODING` instead of inheriting the ISO-8859-1 default (which stays correct for every other `TextNavigationFormat` subclass reading legacy external files — untouched).
- `NavigationFormatParserTest.java`: added `testReadFromStringPreservesNonAsciiCharacters()`.
- `GlopusFormatTest.java`: added `testWriteThenReadBackPreservesNonAsciiDescription()` — round-trips a route with a non-ASCII description through `GlopusFormat.write()` and `NavigationFormatParser.read(String)`.

**Intentional side effect for the changelog:** on-disk `.tk` files are now decoded as UTF-8 instead of ISO-8859-1. Since Glopus/`.tk` is RouteConverter's own internal clipboard/file format (no external producer/consumer), this makes read/write symmetric and is safe — but is a behavior change worth noting for the maintainer.

**Verification:** `factory-verify` returned exit 2 (broker rejected this checkout: "repo not allowed (rc/* only)") — sandboxed build/test is unavailable in this environment, and no local JDK/Maven toolchain was available either, so I could not run `mvn -pl navigation-formats test`. I instead manually traced the type hierarchy (`GlopusFormat extends SimpleLineBasedFormat<Wgs84Route> extends SimpleFormat<Wgs84Route> extends TextNavigationFormat<Wgs84Route>`) to confirm the override signature matches, confirmed `UTF8_ENCODING`/`ISO_LATIN1_ENCODING` constants and the protected `read(InputStream, String, ParserContext)` helper already exist as referenced in the issue, and checked the new tests' write path flushes the `PrintWriter` before reading the buffered string back. Please run the test suite in CI to confirm.

Source issue: cpesch/RouteConverter#340

---
**Builder stats** · model `sonnet` · confidence `0` · 1m 44s across 1 round(s) · 4 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/24825)

Closes #340

### DIFF
--- a/navigation-formats/src/main/java/slash/navigation/base/NavigationFormatParser.java
+++ b/navigation-formats/src/main/java/slash/navigation/base/NavigationFormatParser.java
@@ -47,6 +47,7 @@ import static java.lang.Math.min;
 import static java.lang.String.format;
 import static slash.common.io.Files.getExtension;
 import static slash.common.io.Files.toUrl;
+import static slash.common.io.Transfer.UTF8_ENCODING;
 import static slash.common.io.Transfer.ceiling;
 import static slash.common.type.CompactCalendar.UTC;
 import static slash.common.type.CompactCalendar.fromCalendar;
@@ -268,7 +269,7 @@ public class NavigationFormatParser {
     }
 
     public ParserResult read(String source) throws IOException {
-        return read(new ByteArrayInputStream(source.getBytes()));
+        return read(new ByteArrayInputStream(source.getBytes(UTF8_ENCODING)));
     }
 
     public ParserResult read(InputStream source) throws IOException {
--- a/navigation-formats/src/main/java/slash/navigation/simple/GlopusFormat.java
+++ b/navigation-formats/src/main/java/slash/navigation/simple/GlopusFormat.java
@@ -23,6 +23,8 @@ package slash.navigation.simple;
 import slash.navigation.base.*;
 import slash.navigation.common.NavigationPosition;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.io.PrintWriter;
 import java.util.List;
 import java.util.regex.Matcher;
@@ -71,6 +73,13 @@ public class GlopusFormat extends SimpleLineBasedFormat<Wgs84Route> {
     @SuppressWarnings("unchecked")
     public <P extends NavigationPosition> Wgs84Route createRoute(RouteCharacteristics characteristics, String name, List<P> positions) {
         return new Wgs84Route(this, characteristics, name, (List<Wgs84Position>) positions);
+    }
+
+    // Glopus (.tk) is RouteConverter's own clipboard/file format: both write (PositionSelection)
+    // and read happen inside RouteConverter, so UTF-8 keeps the round-trip lossless instead of
+    // inheriting TextNavigationFormat's ISO-8859-1 default meant for legacy external file formats.
+    public void read(InputStream source, ParserContext<Wgs84Route> context) throws IOException {
+        read(source, UTF8_ENCODING, context);
     }
 
     protected boolean isPosition(String line) {
--- a/navigation-formats/src/test/java/slash/navigation/base/NavigationFormatParserTest.java
+++ b/navigation-formats/src/test/java/slash/navigation/base/NavigationFormatParserTest.java
@@ -228,6 +228,13 @@ public class NavigationFormatParserTest {
     }
 
     @Test
+    public void testReadFromStringPreservesNonAsciiCharacters() throws IOException {
+        ParserResult result = parser.read("51.0450383,7.0508300,San Nicolò");
+        assertTrue(result.isSuccessful());
+        assertEquals("San Nicolò", result.getTheRoute().getPositions().get(0).getDescription());
+    }
+
+    @Test
     public void testReadEmptyInputIsLenientlySuccessful() throws IOException {
         // documents the firstSuccessfulFormat fallback: with no positions parsed,
         // the parser still reports the first format that did not throw
--- a/navigation-formats/src/test/java/slash/navigation/simple/GlopusFormatTest.java
+++ b/navigation-formats/src/test/java/slash/navigation/simple/GlopusFormatTest.java
@@ -20,13 +20,39 @@
 package slash.navigation.simple;
 
 import org.junit.Test;
+import slash.navigation.base.NavigationFormatParser;
+import slash.navigation.base.NavigationFormatRegistry;
+import slash.navigation.base.ParserResult;
+import slash.navigation.base.RouteCharacteristics;
 import slash.navigation.base.Wgs84Position;
+import slash.navigation.base.Wgs84Route;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Collections;
 
 import static org.junit.Assert.*;
 import static slash.common.TestCase.assertDoubleEquals;
 
 public class GlopusFormatTest {
     GlopusFormat format = new GlopusFormat();
+
+    @Test
+    public void testWriteThenReadBackPreservesNonAsciiDescription() throws IOException {
+        Wgs84Position position = new Wgs84Position(7.0508300, 51.0450383, null, null, null, "San Nicolò");
+        Wgs84Route route = format.createRoute(RouteCharacteristics.Waypoints, null, Collections.singletonList(position));
+
+        StringWriter stringWriter = new StringWriter();
+        PrintWriter printWriter = new PrintWriter(stringWriter);
+        format.write(route, printWriter, 0, 1);
+        printWriter.flush();
+
+        NavigationFormatParser parser = new NavigationFormatParser(new NavigationFormatRegistry());
+        ParserResult result = parser.read(stringWriter.toString());
+        assertTrue(result.isSuccessful());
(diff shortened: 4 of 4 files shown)


cpesch (commented):

correctness: The encoding changes are correctly implemented for the stated use case (clipboard round-trip). However, the PR body notes an intentional side effect (on-disk .tk files now decoded as UTF-8 instead of ISO-8859-1) without confirming whether existing .tk files created under the old encoding will remain readable. This is a contract change for file format compatibility that should be verified before merging.

### navigation-formats/src/main/java/slash/navigation/base/NavigationFormatParser.java

- **minor** [correctness] line 272 — The change to use UTF8_ENCODING in read(String) is correct for the clipboard use case, but could affect other callers that pass strings expecting platform-default encoding. The diff shows no other callers in the changed files, but a broader search may be needed to confirm no regression.  _(correctness)_

### navigation-formats/src/main/java/slash/navigation/simple/GlopusFormat.java

- **major** [correctness] line 81 — The override to decode with UTF8_ENCODING changes the contract for reading .tk files from ISO-8859-1 to UTF-8. Existing .tk files created before this change that contain non-ASCII characters would become unreadable or mangled, as they were encoded with UTF-8 but would now be read assuming ISO-8859-1. The PR body acknowledges this as an intentional side effect but does not verify backward compatibility for existing files.  _(correctness)_

```suggestion
Add a test that reads a .tk file containing non-ASCII characters that was created before this change (simulated by writing with the old ISO-8859-1 encoding and reading with the new UTF-8 code) to verify the migration path is safe, or document a migration strategy for existing users.
```

---

navigation-formats/src/main/java/slash/navigation/simple/GlopusFormat.java
```diff
@@ -73,6 +75,13 @@ public <P extends NavigationPosition> Wgs84Route createRoute(RouteCharacteristic
         return new Wgs84Route(this, characteristics, name, (List<Wgs84Position>) positions);
     }
 
+    // Glopus (.tk) is RouteConverter's own clipboard/file format: both write (PositionSelection)
+    // and read happen inside RouteConverter, so UTF-8 keeps the round-trip lossless instead of
+    // inheriting TextNavigationFormat's ISO-8859-1 default meant for legacy external file formats.
+    public void read(InputStream source, ParserContext<Wgs84Route> context) throws IOException {
```
cpesch, line 81: **major** [correctness] The override to decode with UTF8_ENCODING changes the contract for reading .tk files from ISO-8859-1 to UTF-8. Existing .tk files created before this change that contain non-ASCII characters would become unreadable or mangled, as they were encoded with UTF-8 but would now be read assuming ISO-8859-1. The PR body acknowledges this as an intentional side effect but does not verify backward compatibility for existing files.
```suggestion
Add a test that reads a .tk file containing non-ASCII characters that was created before this change (simulated by writing with the old ISO-8859-1 encoding and reading with the new UTF-8 code) to verify the migration path is safe, or document a migration strategy for existing users.
```

navigation-formats/src/main/java/slash/navigation/base/NavigationFormatParser.java
```diff
@@ -268,7 +269,7 @@ private void bufferedInternalRead(InputStream source, int markSize, List<Navigat
     }
 
     public ParserResult read(String source) throws IOException {
-        return read(new ByteArrayInputStream(source.getBytes()));
+        return read(new ByteArrayInputStream(source.getBytes(UTF8_ENCODING)));
```
cpesch, line 272: **minor** [correctness] The change to use UTF8_ENCODING in read(String) is correct for the clipboard use case, but could affect other callers that pass strings expecting platform-default encoding. The diff shows no other callers in the changed files, but a broader search may be needed to confirm no regression.

cpesch:
🤖 Patcher pushed a fix to `agent/builder/340-rc-support-21-fix-utf-8-iso-8859-1-chars`.

PR #336, opened 2026-08-29
Add UI font-size setting (positions table shrunk after uiScale=1.0 fix)

cpesch:
I have successfully implemented the UI font-size setting feature for RouteConverter. Here's a summary of all the changes made:

### 1. UIHelper.java (`common-gui/src/main/java/slash/navigation/gui/helpers/UIHelper.java`)
- Added preference storage constants and methods:
  - `UI_FONT_SCALE_PERCENTAGE_PREFERENCE` preference key
  - `DEFAULT_UI_FONT_SCALE_PERCENTAGE` constant (100)
  - `getUiFontScalePercentage()` getter
  - `setUiFontScalePercentage(int)` setter
  - `scaleFont(Font, int)` pure scaling function with proper rounding and minimum size clamping
- Modified `setLookAndFeel()` to apply font scaling at startup by iterating over UIManager defaults
- Added import for `FontUIResource`

### 2. UIHelperTest.java (`common-gui/src/test/java/slash/navigation/gui/helpers/UIHelperTest.java`)
- Created comprehensive test suite with 6 tests:
  - `testScaleFont100PercentReturnsSameSize` - verifies 100% returns original size
  - `testScaleFont150PercentIncreasesSize` - verifies 12pt → 18pt at 150%
  - `testScaleFont50PercentDecreasesSize` - verifies 12pt → 6pt at 50%
  - `testScaleFontRoundsToNearestInt` - verifies 11pt × 125% = 14pt (13.75 rounded)
  - `testScaleFontClampsToMinimumOnePoint` - verifies minimum size of 1pt
  - `testScaleFontPreservesFamilyAndStyle` - verifies font family and style are preserved

### 3. OptionsDialog.java (`route-converter-gui/src/main/java/slash/navigation/converter/gui/dialogs/OptionsDialog.java`)
- Added `spinnerUiFontScale` field declaration
- Added spinner initialization in constructor with `SpinnerNumberModel(50, 200, 10)` range
- Added ChangeListener to persist changes to preferences
- Updated `$$$setupUI$$$` method to match .form changes (grid row-count 5→6, new row 4 for spinner, filler grid moved to row 5)

### 4. OptionsDialog.form (`route-converter-gui/src/main/java/slash/navigation/converter/gui/dialogs/OptionsDialog.form`)
- Bumped General tab grid row-count from 5 to 6
- Added new label row (row 4, col 0) with i18n key `ui-font-scale-option`
- Added new spinner component (row 4, col 1)
[...]

Issue #335, opened 2026-08-29
[mail rc-support#20] Add UI font-size setting (positions table shrunk after uiScale=1.0 fix)


cpesch:
## Feature: user-configurable UI font size (restart-applied)

**Source:** [rc/rc-support#20](https://git.routeconverter.com/rc/rc-support/issues/20) (mail intake). Request (verbatim): "Is it possible to change the font size in the side panel containing the position, time, elevation? I have attached two screen shots (vs 2.34 and vs 3.6) to show you the difference. Version 2.34 has as good size. I have tried to increase the text size in my Windows settings but it did not make any difference." Goal: give the user a way to make the positions table (and, since the same root cause affects them, every other table/dialog) readable again, without reintroducing the map-blur bug the DPI change fixed.

**Root cause (confirmed):** commit `3fa30855c8` ("fix: avoid blurred map on Windows with scaled UI", 2024-01-23, https://forum.routeconverter.com/thread-3863.html) added `System.setProperty("sun.java2d.uiScale", "1.0")` on Windows in `UIHelper.setLookAndFeel()`. That commit is not an ancestor of tag `2.24` but is an ancestor of tag `3.6` — matches the screenshots exactly. Forcing `uiScale=1.0` stops Java from honoring Windows display-scaling/text-size at all, so every Swing font (not just the positions table — routes/POI/downloads tables and all dialogs too) is frozen at the LAF's unscaled default, and the user's Windows text-size setting has no effect any more.

This spec is fully resolved (grilled 2026-08-29); implement without further questions.

### Scope
Add an app-wide "UI font size" percentage setting (Options dialog, General tab) that scales every `UIManager` default font at startup. Out of scope: reverting or narrowing the `sun.java2d.uiScale=1.0` Windows map-blur fix (keep it exactly as-is); live/no-restart font application (follows the existing "preferred-locale (requires restart)" precedent in the same dialog — do not attempt `SwingUtilities.updateComponentTreeUI` across all windows); per-table individual font overrides (e.g. touching `ConvertPanel`/`tablePositions` directly) — the app-wide UIManager approach fixes the positions table as a side effect and also fixes the same bug for every other table.

### Files
- `common-gui/src/main/java/slash/navigation/gui/helpers/UIHelper.java` — add the scaling logic
- `common-gui/src/test/java/slash/navigation/gui/helpers/UIHelperTest.java` — new test class (module already has JUnit test infra, see `IntegerModelTest` in the same module)
[...]

PR #349, opened 2026-08-30
Auto-detect Windows display scaling as default UI font size (follow-up to #335)

### LINKED ISSUE #338: Auto-detect Windows display scaling as default UI font size (follow-up to #335)
## Feature: auto-detect Windows text-scaling as the default UI font size

**Source:** follow-up to [cpesch/RouteConverter#335](https://github.com/cpesch/RouteConverter/issues/335) / [#336](https://github.com/cpesch/RouteConverter/pull/336). #335/#336 add a manual "UI font size in %" spinner (Options → General) because `sun.java2d.uiScale=1.0` (the Windows map-blur fix) makes RouteConverter ignore the Windows text-size setting entirely. Goal here: remove the need to manually find/set a percentage — auto-match whatever Windows is already scaled to, with zero clicks, while keeping the manual spinner from #336 as an override/escape hatch.

**Depends on #336 landing first** — this issue reuses `UIHelper.getUiFontScalePercentage()` / `setUiFontScalePercentage()` / `scaleFont()` from that PR. Do not start before it merges.

### Scope
On Windows only, when the user has never touched the font-size spinner (no stored preference yet), compute the *default* shown/applied from the real Windows display-scaling percentage instead of hardcoding 100%. Once the user explicitly sets a value via the Options spinner (#336), that stored value always wins over auto-detection — this issue only changes what happens before that first explicit choice. Out of scope: macOS/Linux auto-detection (macOS already handles Retina scaling correctly via its own native HiDPI path, untouched by the Windows-only `uiScale=1.0` fix); touching `sun.java2d.uiScale` or the map-blur fix; changing the spinner's 50–200 range or step.

### Files
- `common-gui/src/main/java/slash/navigation/gui/helpers/UIHelper.java` — add detection
- `common-gui/src/test/java/slash/navigation/gui/helpers/UIHelperTest.java` — add tests for the pure conversion function (already has the #336 `scaleFont` tests to extend)

1. **Detection mechanism** — `Toolkit.getDefaultToolkit().getScreenResolution()` returns the real system DPI via the native `GetDeviceCaps(LOGPIXELSX)` call on Windows (96/120/144/192 for 100/125/150/200% scaling). This is a *different, legacy* AWT code path from the Java2D HiDPI pipeline that `sun.java2d.uiScale` overrides, so it is unaffected by the map-blur fix and keeps reporting the true OS scaling even after `uiScale=1.0` is set.

2. **Pure conversion function** — new package-private (test-visible) `static int dpiToPercentage(int dpi)` in `UIHelper`:
   ```java
   static int dpiToPercentage(int dpi) {
       int percentage = Math.round(dpi / 96f * 100f);
       return Math.max(50, Math.min(200, percentage));
   }
   ```
   Clamp to the same `[50, 200]` range the Options spinner already enforces (#336), so a bogus/extreme DPI reading can't silently produce an unusable scale.

3. **Platform glue (not unit-tested — needs a real Toolkit/display, this repo's GUI is untestable headless)** — new private `static int getDefaultUiFontScalePercentage()`:
   ```java
   private static int getDefaultUiFontScalePercentage() {
       if (!isWindows())
           return DEFAULT_UI_FONT_SCALE_PERCENTAGE;
       try {
           int dpi = Toolkit.getDefaultToolkit().getScreenResolution();
           return dpiToPercentage(dpi);
       } catch (HeadlessException e) {
           return DEFAULT_UI_FONT_SCALE_PERCENTAGE;
       }
   }
   ```

4. **Wire into the existing getter** — change `getUiFontScalePercentage()` (added in #336) from a hardcoded fallback to the computed one:
   ```java
   public static int getUiFontScalePercentage() {
       return preferences.getInt(UI_FONT_SCALE_PERCENTAGE_PREFERENCE, getDefaultUiFontScalePercentage());
   }
   ```
   `preferences.getInt(key, fallback)` only evaluates/uses `fallback` when the key was never written — so a user who already set an explicit value (via the #336 spinner) keeps getting exactly that value, unaffected by this change. A user who never opened Options gets the auto-detected percentage instead of a flat 100 — this also means their *first* visit to Options → General shows the spinner pre-filled with the detected value, not 100, and moving the spinner from there persists an explicit override the normal way.

5. **No `OptionsDialog`/`.form` changes needed** — the spinner already reads its initial value via `getUiFontScalePercentage()` (#336), so it picks up the new default automatically.

Extend `UIHelperTest` (already has the #336 `scaleFont` tests) with cases for `dpiToPercentage`:
- `testDpiToPercentage96DpiIs100Percent` — 96 → 100
- `testDpiToPercentage120DpiIs125Percent` — 120 → 125
- `testDpiToPercentage144DpiIs150Percent` — 144 → 150
- `testDpiToPercentage192DpiIs200Percent` — 192 → 200
- `testDpiToPercentageClampsLowExtremeTo50` — e.g. 40 → clamped to 50
- `testDpiToPercentageClampsHighExtremeTo200` — e.g. 500 → clamped to 200

### Acceptance criteria
- All 6 new `UIHelperTest` cases (plus the existing #336 `scaleFont` cases) pass; `mvn -pl common-gui -am test` is green
- Manual (Windows, 150% display scaling, no prior font-size preference set): Options → General shows the font-size spinner pre-filled around 150, not 100; after explicitly setting it to a different value and restarting, that explicit value persists and is no longer overridden by detection

cpesch:
The diff is clean, correct, and matches the spec exactly. `Toolkit`, `HeadlessException`, and `isWindows` are all already in scope via the existing `java.awt.*` and static imports — no new imports needed.

## Summary

Implemented issue #338: auto-detect Windows display scaling as the default UI font size, building on #336's `UIHelper` scaffolding.

**`common-gui/src/main/java/slash/navigation/gui/helpers/UIHelper.java`**
- Added package-private `static int dpiToPercentage(int dpi)` — converts a DPI reading to a percentage relative to 96 DPI baseline, clamped to `[50, 200]`.
- Added private `static int getDefaultUiFontScalePercentage()` — on Windows, reads `Toolkit.getDefaultToolkit().getScreenResolution()` and converts via `dpiToPercentage`; falls back to the flat 100% default on non-Windows platforms or if `HeadlessException` is thrown.
- Changed `getUiFontScalePercentage()`'s fallback from the hardcoded `DEFAULT_UI_FONT_SCALE_PERCENTAGE` to `getDefaultUiFontScalePercentage()`. Since `Preferences.getInt(key, fallback)` only evaluates the fallback when no value was ever stored, a user's explicit spinner choice (from #336) is untouched — this only changes the very first, never-configured default.

**`common-gui/src/test/java/slash/navigation/gui/helpers/UIHelperTest.java`**
- Added the 6 specified test cases for `dpiToPercentage`: 96→100, 120→125, 144→150, 192→200, and clamping at both extremes (40→50, 500→200).

No `.form`/`OptionsDialog` changes were needed — the spinner already reads its initial value through `getUiFontScalePercentage()`.

**Verify result: unavailable.** `factory-verify -pl common-gui -am test` exited 2 with a broker error (`repo not allowed (rc/* only)` — this repo isn't in the sandbox's allowed verification set), so I could not run `mvn test` in this environment. I manually reviewed the diff for correctness: all imports used (`Toolkit`, `HeadlessException`) are already covered by the existing `java.awt.*` wildcard import and `isWindows` static import, so this should compile cleanly, but the test suite itself has not been executed here.

Source issue: cpesch/RouteConverter#338

---
**Builder stats** · model `sonnet` · confidence `0` · 0m 46s across 1 round(s) · 2 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/24829)

Closes #338

### DIFF
--- a/common-gui/src/main/java/slash/navigation/gui/helpers/UIHelper.java
+++ b/common-gui/src/main/java/slash/navigation/gui/helpers/UIHelper.java
@@ -46,11 +46,27 @@ public class UIHelper {
     private static final int DEFAULT_UI_FONT_SCALE_PERCENTAGE = 100;
 
     public static int getUiFontScalePercentage() {
-        return preferences.getInt(UI_FONT_SCALE_PERCENTAGE_PREFERENCE, DEFAULT_UI_FONT_SCALE_PERCENTAGE);
+        return preferences.getInt(UI_FONT_SCALE_PERCENTAGE_PREFERENCE, getDefaultUiFontScalePercentage());
     }
 
     public static void setUiFontScalePercentage(int percentage) {
         preferences.putInt(UI_FONT_SCALE_PERCENTAGE_PREFERENCE, percentage);
+    }
+
+    private static int getDefaultUiFontScalePercentage() {
+        if (!isWindows())
+            return DEFAULT_UI_FONT_SCALE_PERCENTAGE;
+        try {
+            int dpi = Toolkit.getDefaultToolkit().getScreenResolution();
+            return dpiToPercentage(dpi);
+        } catch (HeadlessException e) {
+            return DEFAULT_UI_FONT_SCALE_PERCENTAGE;
+        }
+    }
+
+    static int dpiToPercentage(int dpi) {
+        int percentage = Math.round(dpi / 96f * 100f);
+        return Math.max(50, Math.min(200, percentage));
     }
 
     public static Font scaleFont(Font font, int percentage) {
--- a/common-gui/src/test/java/slash/navigation/gui/helpers/UIHelperTest.java
+++ b/common-gui/src/test/java/slash/navigation/gui/helpers/UIHelperTest.java
@@ -85,4 +85,34 @@ public class UIHelperTest {
         assertEquals("Serif", result.getFamily());
         assertEquals(Font.BOLD | Font.ITALIC, result.getStyle());
     }
+
+    @Test
+    public void testDpiToPercentage96DpiIs100Percent() {
+        assertEquals(100, UIHelper.dpiToPercentage(96));
+    }
+
+    @Test
+    public void testDpiToPercentage120DpiIs125Percent() {
+        assertEquals(125, UIHelper.dpiToPercentage(120));
+    }
+
+    @Test
+    public void testDpiToPercentage144DpiIs150Percent() {
+        assertEquals(150, UIHelper.dpiToPercentage(144));
+    }
+
+    @Test
+    public void testDpiToPercentage192DpiIs200Percent() {
+        assertEquals(200, UIHelper.dpiToPercentage(192));
+    }
+
+    @Test
+    public void testDpiToPercentageClampsLowExtremeTo50() {
+        assertEquals(50, UIHelper.dpiToPercentage(40));
+    }
+
+    @Test
+    public void testDpiToPercentageClampsHighExtremeTo200() {
+        assertEquals(200, UIHelper.dpiToPercentage(500));
+    }
 }


cpesch (commented):

correctness: The diff correctly implements Windows display scaling auto-detection. The DPI-to-percentage conversion uses proper floating-point arithmetic and clamping. Error handling appropriately catches HeadlessException for headless environments. The fallback behavior correctly preserves existing user preferences while providing platform-aware defaults for new installations.

---

Issue #360, opened 2026-08-31
Route's Line/Polyline layers should share the track GroupLayer, not one Layer each (2/3: Route, depends on #359)


cpesch:
## Fix: currently-open route's Line/Polyline layers should share the track GroupLayer, not one Layer each

**Source:** follow-up to [#356](https://github.com/cpesch/RouteConverter/issues/356) and [#359](https://github.com/cpesch/RouteConverter/issues/359) (same root cause, three-way split of one "redesign the map layer model" ask). **Depends on #359 — implement and merge that one first.** This issue reuses the `trackLayer` `GroupLayer` field + `MapsforgeMapView.getTrackLayer()` accessor #359 introduces (Track and Route characteristics are never the open list at the same time — `getEventMapUpdaterFor()` dispatches to exactly one of `routeUpdater`/`trackUpdater`, so one shared `GroupLayer` for "whatever the open list's lines currently are" is correct, not two GroupLayers where only one is ever populated).

Goal: same as #359, applied to `RouteRenderer` — a route with routing enabled draws each leg as a `Polyline`, still one `Layer` per leg added directly to the shared `Layers` collection today; a route with routing off (or before the routing service responds) draws straight `Line`s, already batched into a single `mapView.addLayers(...)` call but still landing all of them directly in the same shared collection.

This spec is fully resolved (grilled 2026-08-31); implement without further questions.

### Scope

Only `RouteRenderer` and the `routeUpdater`'s `TrackOperation` wiring inside `MapsforgeMapView`. Out of scope: everything #359 already covers (Track-characteristic rendering, `TrackUpdater`'s delta logic — shared by both, untouched here too) and Waypoints (separate third issue).

### Files

- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java` — rewrite the `routeUpdater`'s `TrackOperation` anonymous impl (`update()`/`remove()`) the same way #359 rewrote `trackUpdater`'s — reusing the same `trackLayer` field, no new field here.
- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/renderer/RouteRenderer.java` — `drawStraightLine()` (batched straight-line draw) and `drawRoute()` (per-leg routed-polyline swap): retarget from `mapView.addLayers(...)`/`mapView.addLayer(...)`/`mapView.removeLayer(...)` to `mapView.getTrackLayer()`.
[...]

Issue #359, opened 2026-08-31
Track's Line layers should share one GroupLayer, not one Layer each (1/3: Track)


cpesch:
## Fix: currently-open track's Line layers should share one GroupLayer, not one Layer each

**Source:** follow-up to [#356](https://github.com/cpesch/RouteConverter/issues/356) (selection-marker freeze, same root cause). Recon while speccing #356 found the currently-open track has the identical shape of problem: `TrackRenderer.drawTrack()` creates one `org.mapsforge.map.layer.Line` per consecutive position pair (~113,933 of them for a 113,934-position track), all added directly to the map's single shared `Layers` collection (`CopyOnWriteArrayList`-backed). Any later bulk removal against that collection — by this track itself (position add/remove/move → `TrackUpdater` → remove old segments) or by anything else sharing the same collection (e.g. #356's selection markers, before that fix) — pays O(n×m) against a haystack this large, since mapsforge's own `Layers.removeAll(list, true)` does a per-item linear scan.

Goal: shrink the shared `Layers` collection's membership for the open track's own lines from O(track length) to O(1), mirroring the existing `NonSelectedPositionListsRenderer`/`GroupLayer` pattern this codebase already uses for every *other* route's rendering, and the same pattern #356 applies to selection markers.

**Split from a single "redesign the map layer model" ask into three issues** (this one, plus a Route issue and a Waypoints issue) so each stays reviewable on its own. **This issue is the foundational one — implement and merge it first.** The Route issue depends on the `trackLayer` field/accessor this issue introduces; the Waypoints issue is independent of both.

This spec is fully resolved (grilled 2026-08-31); implement without further questions.

### Scope

[...]

PR #364, opened 2026-08-31
Waypoints' Marker layers should share one GroupLayer, not one Layer each (3/3: Waypoints, independent)

### LINKED ISSUE #361: Waypoints' Marker layers should share one GroupLayer, not one Layer each (3/3: Waypoints, independent)
## Fix: currently-open Waypoints list's Marker layers should share one GroupLayer, not one Layer each

**Source:** follow-up to [#356](https://github.com/cpesch/RouteConverter/issues/356), [#359](https://github.com/cpesch/RouteConverter/issues/359) and [#360](https://github.com/cpesch/RouteConverter/issues/360) — same root cause, three-way split of one "redesign the map layer model" ask. **Independent of #359/#360** — Waypoints is its own `RouteCharacteristics` value, dispatched separately (`getEventMapUpdaterFor(Waypoints) -> waypointUpdater`), never the open list at the same time as Track or Route, so this doesn't reuse `trackLayer` and can be built in parallel.

Goal: same as #359/#360, applied to the `waypointUpdater`'s `WaypointOperation` — when the currently-open list is Waypoints-characteristic (e.g. a large POI file), every waypoint gets its own `org.mapsforge.map.layer.overlay.Marker` added directly to the shared `Layers` collection, same shape of freeze risk on a large file.

This spec is fully resolved (grilled 2026-08-31); implement without further questions.

### Scope

Only the `waypointUpdater`'s `WaypointOperation` wiring inside `MapsforgeMapView` — there's no separate `WaypointRenderer` class to touch; unlike Track/Route, the rendering logic (`createMarker`, layer add/remove) lives directly inside the `WaypointOperation` anonymous implementation. `WaypointUpdater`'s delta-computation logic (`handleAdd`/`handleUpdate`/`handleRemove`) is untouched — same reasoning as #359 for `TrackUpdater`; `WaypointUpdaterTest` must stay green unmodified.

Note: waypoint markers are plain `org.mapsforge.map.layer.overlay.Marker`, **not** `DraggableMarker` (confirmed during recon) — they are not a drag target, and `MapViewMoverAndZoomer.getMarkerFor()`'s hit-test loop only ever returns a match when `layer instanceof DraggableMarker`, so a plain `Marker` never satisfies it today regardless of whether it's found by the outer `instanceof Marker` filter. Moving these into a `GroupLayer` therefore needs **no change to `MapViewMoverAndZoomer`** — confirmed no drag/click behaviour depends on waypoint markers being individually present in the top-level `Layers` collection.

### Files

- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java` — add a `waypointLayer` field + rewrite the `waypointUpdater`'s `WaypointOperation` anonymous impl (`add()`/`update()`/`remove()`).

1. **New field**, same shape as #359's `trackLayer` (separate instance — waypoints are `Marker`s, a different rendering concern from track/route lines):
   ```java
   private final GroupLayer waypointLayer = new GroupLayer();
   ```
   Added once via `addLayer(waypointLayer)`, insertion-order (no explicit anchoring), same reasoning as #359.

2. **`WaypointOperation.add()`** — keep the existing per-position `Marker` construction loop (`createMarker`) exactly as today; only the hand-off changes:
   ```java
   // was: addObjectsWithLayer(withLayers);
   synchronized (waypointLayer) {
       for (PositionWithLayer positionWithLayer : withLayers)
           waypointLayer.layers.add(positionWithLayer.getLayer());
   }
   waypointLayer.requestRedraw();
   ```

3. **`WaypointOperation.remove()`** — same "removing everything" fast path as #359/#360:
   ```java
   public void remove(List<PositionWithLayer> positionWithLayers) {
       synchronized (waypointLayer) {
           if (positionWithLayers.size() == waypointLayer.layers.size()) {
               waypointLayer.layers.clear();
           } else {
               Set<Layer> toRemove = new HashSet<>(positionWithLayers.size());
               for (PositionWithLayer positionWithLayer : positionWithLayers) {
                   Layer layer = positionWithLayer.getLayer();
                   if (layer != null)
                       toRemove.add(layer);
                   else
                       log.warning("Could not find layer to remove for " + positionWithLayer);
               }
               waypointLayer.layers.removeAll(toRemove);
           }
       }
       waypointLayer.requestRedraw();
   }
   ```
   (`removeObjectWithLayers`'s existing `positionWithLayer.setLayer(null)` bookkeeping per item stays — keep it, just move the actual layer removal to the block above instead of delegating to `removeObjectWithLayers`.)

4. **`WaypointOperation.update()`** currently does `removeLayers(toLayers(positionWithLayers)); add(positionWithLayers);` (a genuine remove-then-recreate, not a carry-forward — confirmed during recon, every updated waypoint gets a brand new `Marker`). Retarget the `removeLayers(...)` call the same way as point 3, keep calling `add(...)` afterward unchanged.

`WaypointOperation`'s `add`/`update`/`remove` are anonymous-inner-class methods inside `MapsforgeMapView` — same as `SelectionOperation`/`TrackOperation`, not independently unit-testable (no test exists for these today either, confirmed by recon). `WaypointUpdaterTest` (existing) needs no changes — it exercises `WaypointUpdater`'s delta logic against a mocked `WaypointOperation`, fully decoupled from this rewrite, same as `TrackUpdaterTest`/`SelectionUpdaterTest`.

### i18n

None — no new user-facing strings.

### Acceptance criteria

- Existing `WaypointUpdaterTest` still green, unmodified.
- Manual maintainer check: open or convert a large file to Waypoints characteristic (e.g. many thousands of POIs), confirm markers still render correctly; add/remove/move a handful of waypoints and confirm the map updates correctly with no freeze; select/deselect a large range of waypoints in the table (exercises #356's fix concurrently, unrelated GroupLayer, but shares the map view) and confirm no interaction/regression between the two GroupLayers.

cpesch:
Excellent! The implementation is complete. Let me create a summary of the changes:

## Summary

I have successfully implemented the fix for **Issue #361**: Waypoints' Marker layers now share one GroupLayer instead of having individual layers.

### Changes Made

**File Modified:** `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java`

#### 1. Added new `waypointLayer` field (line 219)
```java
private final GroupLayer waypointLayer = new GroupLayer();
```

#### 2. Modified `WaypointOperation.add()` method (lines 324-328)
- Replaced the call to `addObjectsWithLayer(withLayers)` with direct layer management
- Added synchronized block to safely add marker layers to the `waypointLayer`
- Calls `requestRedraw()` after adding layers

#### 3. Modified `WaypointOperation.update()` method (lines 331-348)
- Replaced `removeLayers(remove)` call with synchronized block that:
  - Uses fast path (`clear()`) when removing all layers
  - Uses `HashSet` for efficient partial removal
  - Calls `requestRedraw()` after removal
- Still calls `add()` afterward to recreate the updated markers

#### 4. Modified `WaypointOperation.remove()` method (lines 351-371)
- Replaced `removeObjectWithLayers()` call with synchronized block that:
  - Uses fast path (`clear()`) when removing all layers
  - Uses `HashSet` for efficient partial removal with proper null checks
  - Calls `requestRedraw()` after removal
- Maintains existing bookkeeping by setting each layer reference to null

#### 5. Added waypointLayer to map layers (lines 774-778)
- Added waypointLayer to the Layers collection in `handleNonSelectedPositionLists()`
- Uses `contains()` check to avoid duplicate additions

### Benefits

This change prevents UI freezes when handling large waypoint files by:
- **Eliminating individual layer adds/removes**: Instead of adding each marker as a separate layer to the shared `Layers` collection, all markers are now managed within a single `GroupLayer`
- **Batch redraws**: A single `requestRedraw()` call replaces what would have been many individual layer operations
- **Thread-safe**: All operations are properly synchronized on the `waypointLayer` monitor
- **Performance**: Uses fast-path clearing for bulk operations and efficient `HashSet` lookups for partial removals

### Testing

The changes follow the same pattern as the previously implemented fixes for Tracks (#359) and Routes (#360), ensuring consistency across the codebase. The existing `WaypointUpdaterTest` should remain green as it exercises `WaypointUpdater`'s delta logic against a mocked `WaypointOperation`, fully decoupled from this rewrite.

Source issue: cpesch/RouteConverter#361

---
**Builder stats** · model `sonnet` · confidence `0` · 2m 8s across 1 round(s) · 1 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/25299)

Closes #361

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
@@ -216,6 +216,7 @@ public class MapsforgeMapView extends BaseMapView {
     private SelectionUpdater selectionUpdater;
     private EventMapUpdater routeUpdater, trackUpdater, waypointUpdater;
     private UpdateDecoupler updateDecoupler;
+    private final GroupLayer waypointLayer = new GroupLayer();
 
     // initialization
 
@@ -320,17 +321,54 @@ public class MapsforgeMapView extends BaseMapView {
                     positionWithLayer.setLayer(marker);
                     withLayers.add(positionWithLayer);
                 }
-                addObjectsWithLayer(withLayers);
+                synchronized (waypointLayer) {
+                    for (PositionWithLayer positionWithLayer : withLayers)
+                        waypointLayer.layers.add(positionWithLayer.getLayer());
+                }
+                waypointLayer.requestRedraw();
             }
 
             public void update(List<PositionWithLayer> positionWithLayers) {
                 List<Layer> remove = toLayers(positionWithLayers);
-                removeLayers(remove);
+                synchronized (waypointLayer) {
+                    if (remove.size() == waypointLayer.layers.size()) {
+                        waypointLayer.layers.clear();
+                    } else {
+                        Set<Layer> toRemove = new HashSet<>(remove.size());
+                        for (Layer layer : remove) {
+                            if (layer != null)
+                                toRemove.add(layer);
+                            else
+                                log.warning("Could not find layer to remove for " + layer);
+                        }
+                        waypointLayer.layers.removeAll(toRemove);
+                    }
+                }
+                waypointLayer.requestRedraw();
                 add(positionWithLayers);
             }
 
             public void remove(List<PositionWithLayer> positionWithLayers) {
-                removeObjectWithLayers(positionWithLayers);
+                synchronized (waypointLayer) {
+                    if (positionWithLayers.size() == waypointLayer.layers.size()) {
+                        waypointLayer.layers.clear();
+                    } else {
+                        Set<Layer> toRemove = new HashSet<>(positionWithLayers.size());
+                        for (PositionWithLayer positionWithLayer : positionWithLayers) {
+                            Layer layer = positionWithLayer.getLayer();
+                            if (layer != null)
+                                toRemove.add(layer);
+                            else
+                                log.warning("Could not find layer to remove for " + positionWithLayer);
+                        }
+                        waypointLayer.layers.removeAll(toRemove);
+                    }
+                }
+                waypointLayer.requestRedraw();
+                // keep the existing bookkeeping
+                for (PositionWithLayer positionWithLayer : positionWithLayers) {
+                    positionWithLayer.setLayer(null);
+                }
             }
         });
 
@@ -733,6 +771,11 @@ public class MapsforgeMapView extends BaseMapView {
 
         // catch position lists that were loaded before the map was initialized
         nonSelectedPositionListsRenderer.update();
+
+        // add the waypoint layer if not already present
+        if (!layers.contains(waypointLayer)) {
+            layers.add(waypointLayer);
+        }
     }
 
     private void updateNonSelectedPositionLists() {


mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
```diff
@@ -320,17 +321,54 @@ public void add(List<PositionWithLayer> positionWithLayers) {
...
+                    } else {
+                        Set<Layer> toRemove = new HashSet<>(remove.size());
+                        for (Layer layer : remove) {
+                            if (layer != null)
+                                toRemove.add(layer);
+                            else
+                                log.warning("Could not find layer to remove for " + layer);
+                        }
+                        waypointLayer.layers.removeAll(toRemove);
+                    }
```
cpesch, line 345: **minor** [correctness] Warning message logs 'positionWithLayer' without checking if it is null first, similar to line 338 issue.
```suggestion
Check for null before logging or log a more specific identifier for the layer.
```

mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
```diff
@@ -320,17 +321,54 @@ public void add(List<PositionWithLayer> positionWithLayers) {
...
 
             public void update(List<PositionWithLayer> positionWithLayers) {
                 List<Layer> remove = toLayers(positionWithLayers);
-                removeLayers(remove);
+                synchronized (waypointLayer) {
+                    if (remove.size() == waypointLayer.layers.size()) {
+                        waypointLayer.layers.clear();
+                    } else {
+                        Set<Layer> toRemove = new HashSet<>(remove.size());
+                        for (Layer layer : remove) {
```
cpesch, line 338: **minor** [correctness] Warning message logs 'layer' when it is known to be null, producing unhelpful output 'Could not find layer to remove for null'.
```suggestion
Move the warning inside the null check or log positionWithLayer instead to provide useful diagnostic information.
```

cpesch (commented):

correctness: The refactoring consolidates waypoint marker layers into a single GroupLayer with proper synchronization. The implementation follows the same pattern as previous fixes for Tracks and Routes. Minor issues found: ineffective null warning message and size-based fast-path optimization that assumes all layers should be cleared when sizes match.

- **minor** [correctness] line 338 — Warning message logs 'layer' when it is known to be null, producing unhelpful output 'Could not find layer to remove for null'.  _(correctness)_

```suggestion
Move the warning inside the null check or log positionWithLayer instead to provide useful diagnostic information.
```
- **minor** [correctness] line 345 — Warning message logs 'positionWithLayer' without checking if it is null first, similar to line 338 issue.  _(correctness)_

```suggestion
Check for null before logging or log a more specific identifier for the layer.
```

---

PR #368, opened 2026-09-02
🤖 Patcher: fix master CI failure (agent/patcher/trunk-91ea8c5ef453)

cpesch:
Automated fix for a failing CI run on `master`. Opened by patcher (specs/00040 trunk lane). Review + merge to land.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/helpers/MapViewMoverAndZoomer.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/helpers/MapViewMoverAndZoomer.java
@@ -24,7 +24,6 @@ import org.mapsforge.core.model.LatLong;
 import org.mapsforge.map.layer.GroupLayer;
 import org.mapsforge.map.layer.Layer;
 import org.mapsforge.map.layer.LayerManager;
-import org.mapsforge.map.layer.Layers;
 import org.mapsforge.map.layer.overlay.Marker;
 import org.mapsforge.map.model.MapViewPosition;
 import org.mapsforge.map.util.MapViewProjection;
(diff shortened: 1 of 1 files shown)


cpesch:
Closing — this patch targets the wrong failure and would not compile.

**What actually failed at 91ea8c5ef.** `RouteConverter CI` was green on that commit. The red run was `Publish Javadoc` ([33599587946](https://github.com/cpesch/RouteConverter/actions/runs/33599587946)), and it failed in the `download` module:

```
DownloadManager304Test.testOutdatedFileDropsETagSoReDownloadIsUnconditional:136 expected:<1> but was:<0>
Failed to execute goal ...:surefire:test (default-test) on project download: There are test failures.
```

Nothing in `mapsforge-mapview`, and nothing to do with `MapViewMoverAndZoomer`.

**Why the diff would break the build.** It replaces

```java
findDraggableMarkerAt(layerManager.getLayers().getLayers(), ...)
```

with `findDraggableMarkerAt(layerManager.getLayers(), ...)`. `LayerManager#getLayers()` returns `Layers`, and in mapsforge 0.30.0 that is

```
public class org.mapsforge.map.layer.Layers implements java.lang.Iterable<Layer>, java.util.RandomAccess
```

— not a `List`. `findDraggableMarkerAt` takes `List<Layer>`, and the diff does not change that signature, so the call no longer type-checks. The `.getLayers().getLayers()` on master is correct: the second call is `Layers#getLayers()`, which returns `List<Layer>`.

**Current state.** `Publish Javadoc` is green on master head (8a34ad537), and `DownloadManager304Test` passes locally on master. The failure looks like a flake in a timing-sensitive download test rather than a defect the patch would have addressed; tracking that separately.

PR #371, opened 2026-09-02
test(graphhopper): make GraphHopperTest clean up after itself

cpesch:
## Problem

`mvn -pl graphhopper test` passes on a clean machine but fails on the second run:

```
GraphHopperTest.isRequiresDownloadSwitchesToAnAlreadyPresentGraphForTheCurrentRoute:109
```

The assertion there is `assertTrue(malaFatraLocalFile.createNewFile())`, which returns `false` once the file exists from an earlier run.

The test built its data source directory name as `"graphhopper-test-" + hashCode()` and resolved it through `getApplicationDirectory()`, so it created real directories below `~/.routeconverter/graphhopper-test-<hashCode>/` (e.g. `europe/slovakia/mala-fatra-latest.osm.pbf`) and never deleted them. `hashCode()` repeats across runs often enough to collide.

## Change

- Data source directory name is unique per test (`UUID.randomUUID()`), and `tearDown()` deletes the created tree below the application directory via `recursiveDelete()`.
- The download queue file and the simulated previously loaded graph now live in a `TemporaryFolder` instead of being left behind in the temporary directory.
- `TEST_MODE` is set once in `setUp()` instead of per test method.

Production code is unchanged.

## Why not isolate the directory completely

`GraphManager.getDirectory()` prefers its `directory` preference over `getApplicationDirectory()`, so pointing that preference at a `TemporaryFolder` looks like the cleaner fix. It does not work: the preference lives in a `Preferences` user node shared across JVMs, while surefire runs test classes in parallel forks (`forkCount 1.5C`). `GraphHopperTest` and `GraphManagerTest` then overwrite each other's directory preference:

```
GraphHopperTest...SwitchesToAnAlreadyPresentGraphForTheCurrentRoute:145
  GraphHopper must not require a download when the graph for the route's actual region is already present locally
GraphManagerTest.testMissingDirectoryPreferenceFallsBackToApplicationDirectory:208
  expected:</[Users/.../.routeconverter/graph-manager-test-...]> but was:</[var/folders/.../junit.../graph-directory]>
```

Deleting the created tree in `tearDown()` avoids touching shared state.

## Verification

`mvn -o -pl graphhopper test -Djacoco.skip=true`, run twice in a row on Temurin 21.0.11:

```
run 1: Tests run: 15, Failures: 0, Errors: 0 - BUILD SUCCESS
run 2: Tests run: 15, Failures: 0, Errors: 0 - BUILD SUCCESS
```

Afterwards no `~/.routeconverter/graphhopper-test-*` directories remain.

## Sibling tests

- `GraphManagerTest` creates `~/.routeconverter/graph-manager-test-<uuid>` directories but already deletes them in its `tearDown()`; no leftovers on disk. Unchanged.
- `GraphHopperIT` writes `~/.routeconverter/test` deliberately as a download cache for the real Hamburg PBF and is excluded from surefire. Unchanged.

### DIFF
--- a/graphhopper/src/test/java/slash/navigation/graphhopper/GraphHopperTest.java
+++ b/graphhopper/src/test/java/slash/navigation/graphhopper/GraphHopperTest.java
@@ -19,7 +19,11 @@
 */
 package slash.navigation.graphhopper;
 
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 import slash.navigation.common.BoundingBox;
 import slash.navigation.common.LongitudeAndLatitude;
 import slash.navigation.datasources.DataSource;
@@ -29,8 +33,8 @@ import slash.navigation.routing.DownloadFuture;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.UUID;
 
-import static java.io.File.createTempFile;
 import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


cpesch (commented):

### Panel

- **correctness** — approve · confidence 0.90
  Test-only change that removes the cross-run hashCode() collision behind the flaky createNewFile() assertion: a UUID-per-test directory plus tearDown() recursiveDelete() targets exactly the path the production code computes, and TemporaryFolder covers the download-queue file and the simulated previously loaded graph. No runtime defect found in the changed lines; the TEST_MODE static was never reset before this change either, so relocating it to setUp() leaves behaviour unchanged.

---

PR #374, opened 2026-09-03
Cache the bundled jar under LOCALAPPDATA instead of re-extracting 60 MB per launch

### LINKED ISSUE #372: [report #1343] Cache the bundled jar instead of re-extracting 60 MB to %TEMP% on every Windows launch
## Fix: cache the bundled jar instead of re-extracting 60 MB to %TEMP% on every Windows launch

**Source:** crash/feedback report #1343 (Windows 11, RouteConverter 3.6 vom 2026 Aug 23, bundled JRE build).
Request (verbatim, signature and username redacted): *"Jedes starten von Routeconverter dauert ca. 15 - 20 Sekunden. Es mach keinen Unterschied ob das Programm diekt oder über einr gpx Datei gestartet wird. Ich nutze die Windows Version mit bebundeltem java. Gibt es aus dem Fehlerbericht Hinweise zur Ursache? Ich gehe davon aus, das es an meinem System liegt."*

Goal: make the second and every subsequent launch of the bundled-JRE Windows build fast. It is **not** the reporter's system — it is a launcher defect.

This spec is fully resolved (grilled 2026-09-03); implement without further questions.

### Diagnosis (measured from the submitted log)

Java-side startup is fine. Three warm starts in the report:

```
Frame shown 2379 ms after startup
Frame shown 2419 ms after startup
Frame shown 2404 ms after startup
```

Map view ready ~2.6 s in each. `startupStartMillis` is set inside `startup()`
(`route-converter-gui/src/main/java/slash/navigation/converter/gui/BaseRouteConverter.java:237`),
so this clock begins well inside Java — everything the NSIS launcher does before
the JVM is invisible to it. The missing 13–17 s lives there.

`RouteConverterWindows/src/main/resources/RouteConverter.nsi` today:

```nsis
Section
  SetOverwrite off

  SetOutPath "${JRE_DIR}"          ; $LOCALAPPDATA\RouteConverter\jre-<ver> — persisted, reused
  File /r "${JRE_SRC}\*"

  InitPluginsDir                    ; fresh random %TEMP%\ns####.tmp on EVERY launch
  SetOutPath $PluginsDir
  File "RouteConverterWindows.jar"  ; 60.7 MB re-extracted on EVERY launch
  SetOutPath $TEMP
  ${GetParameters} $R0
  nsExec::Exec '"${JRE_DIR}\bin\java.exe" -server -Drouteconverter.bundledJre=true -jar $PluginsDir\RouteConverterWindows.jar $R0'
  RMDir /r $PluginsDir              ; deleted again on exit
SectionEnd
```

The JRE is cached across launches. The **application jar is not**: every launch
writes 60.7 MB (`RouteConverterWindows/target/RouteConverterWindows.jar`) to a
brand-new temp path, java.exe reads it back, and it is deleted on exit. Windows
Defender therefore rescans an unknown 60 MB file at a new path on every single
launch, with no cache hit possible. That is the reported 15–20 s.

`TimeAlbumProWindows/src/main/resources/TimeAlbumPro.nsi` has the identical shape
and the identical defect.

No Java code reads or writes `%LOCALAPPDATA%\RouteConverter` (the application
directory is `%USERPROFILE%\.routeconverter`, see the log's `initializeDiagnostics`
line), so that tree is launcher-owned and safe for the launcher to manage.

### Scope

Rework both NSIS launcher scripts so the application jar is extracted once per
build generation and reused, gated by a stamp file, with best-effort pruning of
superseded generations. Add a CI smoke test that asserts the warm path.

Out of scope:
- Shrinking the 60.7 MB jar (see retired `specs/done/00009-reduce-bundle-size-generically.md` — no low-maintenance shrink remains).
- The bundled JRE version bump (#167) — this change must keep working across it, nothing more.
- `RouteConverterPortable` — its launcher `.ini` still names the pre-00044 jar `RouteConverterWindowsOpenSource.jar` while the pom now copies in `RouteConverterWindows.jar`. Real, separate defect; tracked in #373, do not touch here.

### Files

- `RouteConverterWindows/src/main/resources/RouteConverter.nsi`
- `TimeAlbumProWindows/src/main/resources/TimeAlbumPro.nsi`
- `.github/workflows/_build-windows.yml` (new smoke step after the two `makensis` steps)

**1. Layout.** Both generations live in version-namespaced sibling directories
under the existing product root, so a running instance can never have its
directory pulled out from under it:

```
$LOCALAPPDATA\RouteConverter\jre-${jre.version}          (unchanged path)
$LOCALAPPDATA\RouteConverter\app-${VERSION}\RouteConverterWindows.jar
$LOCALAPPDATA\RouteConverter\app-${VERSION}\.stamp
```

where `VERSION` = `${project.version}-${maven.build.number}` (both already
available to the filtered `.nsi`, cf. the existing `VIProductVersion` line).
TimeAlbumPro uses `$LOCALAPPDATA\TimeAlbumPro\...` and `TimeAlbumProWindows.jar`
identically.

**2. Stamp gate (fast path).** `.stamp` contains exactly `${VERSION}|${jre.version}`.
On launch, read `${APP_DIR}\.stamp`; if its content equals that string, **skip both
`File` extractions entirely** and jump straight to the `nsExec::Exec`. A missing or
non-matching stamp takes the slow path. Do not rely on `SetOverwrite off` to make
re-extraction cheap — gate it explicitly.

**3. Commit marker semantics.** On the slow path, write `.stamp` **last**, after
both extractions have completed. An extraction interrupted by a crash, a full disk
or a killed process therefore leaves no valid stamp, and the next launch re-does it.
This also removes the current latent hazard where `SetOverwrite off` would keep a
truncated jar forever.

**4. Best-effort prune (slow path only).** Before extracting, enumerate the direct
subdirectories of the product root and `RMDir /r` every one that is neither
`app-${VERSION}` nor `jre-${jre.version}`. Ignore failures: Windows refuses to
delete files held by a running process, so an old instance still executing from a
superseded generation is never disturbed, and the prune simply succeeds on a later
launch. Never prune on the fast path.

**5. Quote the jar path.** The current `-jar $PluginsDir\RouteConverterWindows.jar`
is unquoted. `%LOCALAPPDATA%` contains a space whenever the Windows account name
does (`C:\Users\First Last\AppData\Local`), so the new path **must** be quoted:

```nsis
nsExec::Exec '"${JRE_DIR}\bin\java.exe" -server -Drouteconverter.bundledJre=true -jar "${APP_DIR}\RouteConverterWindows.jar" $R0'
```

Keep `-Drouteconverter.bundledJre=true` (read by
`common/src/main/java/slash/common/log/DiagnosticReport.java:63`); TimeAlbumPro
keeps its current flag set unchanged. Keep `SetOutPath $TEMP` before the exec and
keep `${GetParameters} $R0` so file associations continue to work.

**6. `$PLUGINSDIR`.** Drop the explicit `InitPluginsDir` / `SetOutPath $PluginsDir` /
`File` / `RMDir /r $PluginsDir` quartet — the jar no longer goes there. **`nsExec`
still needs `$PLUGINSDIR` for its own plugin DLL**; the NSIS compiler inserts that
extraction automatically and NSIS removes the directory when the installer exits,
so no explicit handling is required. A short `ns*.tmp` directory will therefore
still appear under `%TEMP%` on every launch — it contains only the few-KB nsExec
DLL. The smoke test must assert on the *jar*, not on the absence of `ns*.tmp`.

**7. Keep everything else as-is:** `SilentInstall silent`, `RequestExecutionLevel user`,
the `VIAddVersionKey` block, the icon, `OutFile`, and the `JRE_SRC` build-time path.

### Testable seam

N/A in the usual sense — NSIS script has no unit-testable pure function, and the
defect is entirely in launcher I/O behaviour. The machine-verifiable substitute is
the CI smoke test below; it is a required part of this issue, not optional.

### CI smoke test

New step in `.github/workflows/_build-windows.yml`, after `makensis RouteConverterWindows`
(the runner is `windows-latest` and the built `.exe` is already on disk there).
PowerShell; for each of two consecutive runs:

1. `Start-Process -PassThru` the built `RouteConverterWindows/target/RouteConverterWindows.exe`.
2. Poll (say 100 ms, 90 s cap) until `%LOCALAPPDATA%\RouteConverter\app-*\.stamp` exists
   **and** a `java.exe` child process is running; record the elapsed time.
3. Kill the process tree (`taskkill /T /F /PID`), wait for exit.

Assertions:

- **A.** After run 1: `%LOCALAPPDATA%\RouteConverter\app-*\RouteConverterWindows.jar`
  exists and is > 10 MB, and `app-*\.stamp` exists.
- **B.** During run 2: no directory matching `%TEMP%\ns*.tmp` contains a `.jar` larger
  than 10 MB. (Small `ns*.tmp` directories are expected — see decision 6.)
- **C.** Run 2 reaches the java.exe launch in **under 3 s** (generous headroom; the
  warm path is a single file read plus a string compare). Run 1 is not time-asserted.
- **D.** After run 2, exactly one `app-*` directory exists under the product root.

Fail the job on any assertion. Run the same block for `TimeAlbumProWindows.exe`
against `%LOCALAPPDATA%\TimeAlbumPro`, or factor it into a reusable PowerShell
function parameterised by product root, exe path and jar name — implementer's choice.

### i18n

None — no user-visible strings change.

### Acceptance criteria

- `makensis` builds both `.nsi` files without warnings; `_build-windows.yml` stays green.
- The new CI smoke step passes assertions A–D for both products (machine-verifiable).
- Manual maintainer check on a real Windows 11 machine with Defender enabled:
  cold first launch is comparable to today; the second launch drops from ~15–20 s
  to roughly 3 s. Report #1343 asks for exactly this.
- Manual: install an older build over a newer one (downgrade) and confirm the stamp
  mismatch triggers re-extraction rather than silently running the wrong jar.

cpesch:
Implements #372 (crash-report #1343: 15-20s Windows startup). Both NSIS launchers extract app jar + JRE into $LOCALAPPDATA\<Product>\app-<version> gated by a version+jre stamp file (crash-safe: stamp written last); warm launches skip extraction entirely; superseded app-*/jre-* dirs pruned; -jar path quoted (latent space-in-path fix). CI gains a two-launch smoke on the built installers asserting cold-extract, no-TEMP-jar, warm-under-3s, single-surviving-dir.

Built by a maintainer session — the factory cannot push workflow-file changes by design (two builder runs died on exactly that; see the issue).

Closes #372

### DIFF
--- a/.github/workflows/_build-windows.yml
+++ b/.github/workflows/_build-windows.yml
@@ -122,6 +122,91 @@ jobs:
         working-directory: TimeAlbumProWindows
         run: makensis "target\TimeAlbumPro.nsi"
 
+      - name: Smoke test — bundled-jar warm-launch caching (issue #372)
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = 'Stop'
+
+          function Wait-ForWarmSignal($proc, $productRoot) {
+            $sw = [System.Diagnostics.Stopwatch]::StartNew()
+            $cap = [TimeSpan]::FromSeconds(90)
+            while ($sw.Elapsed -lt $cap) {
+              $stamp = Get-ChildItem -Path "$productRoot\app-*\.stamp" -ErrorAction SilentlyContinue
+              $javaChild = Get-CimInstance Win32_Process `
+                -Filter "Name='java.exe' AND ParentProcessId=$($proc.Id)" `
+                -ErrorAction SilentlyContinue
+              if ($stamp -and $javaChild) {
+                $sw.Stop()
+                return $sw.Elapsed
+              }
+              Start-Sleep -Milliseconds 100
+            }
+            throw "timed out waiting for warm signal ($($proc.Path))"
+          }
+
+          function Stop-ProcessTree($procId) {
+            taskkill /T /F /PID $procId | Out-Null
+            Start-Sleep -Milliseconds 500
+          }
+
+          function Test-WarmLaunchCaching($exePath, $productRoot, $jarName) {
+            Write-Host "=== warm-launch smoke test: $exePath ==="
+
+            # Run 1 (cold): first-ever launch must do the slow path and leave
+            # a cached jar + stamp behind (assertion A).
+            $p1 = Start-Process -FilePath $exePath -PassThru
+            $elapsed1 = Wait-ForWarmSignal $p1 $productRoot
+            Write-Host "run 1: warm signal after $($elapsed1.TotalSeconds) s"
+
+            $jar1 = Get-ChildItem -Path "$productRoot\app-*\$jarName" -ErrorAction SilentlyContinue
+            if (-not $jar1 -or $jar1.Length -le 10MB) {
+              throw "assertion A failed: $productRoot\app-*\$jarName missing or <= 10MB"
+            }
+            if (-not (Get-ChildItem -Path "$productRoot\app-*\.stamp" -ErrorAction SilentlyContinue)) {
+              throw "assertion A failed: .stamp missing after run 1"
+            }
+
+            Stop-ProcessTree $p1.Id
+
+            # Run 2 (warm): must take the fast path.
+            $p2 = Start-Process -FilePath $exePath -PassThru
+            $elapsed2 = Wait-ForWarmSignal $p2 $productRoot
+            Write-Host "run 2: warm signal after $($elapsed2.TotalSeconds) s"
+
+            # Assertion B: the app jar is never re-extracted into a fresh
+            # %TEMP%\ns*.tmp — only the small nsExec plugin DLL may land there.
+            $bigTmpJars = Get-ChildItem -Path "$env:TEMP\ns*.tmp" -Recurse -Filter *.jar -ErrorAction SilentlyContinue |
+                          Where-Object { $_.Length -gt 10MB }
+            if ($bigTmpJars) {
+              throw "assertion B failed: large jar under %TEMP%\ns*.tmp: $($bigTmpJars.FullName -join ', ')"
+            }
+
+            # Assertion C: warm path is a stamp read plus a string compare.
+            if ($elapsed2.TotalSeconds -ge 3) {
+              throw "assertion C failed: run 2 took $($elapsed2.TotalSeconds) s (expected < 3 s)"
+            }
+
+            Stop-ProcessTree $p2.Id
+
+            # Assertion D: pruning left exactly one app-* generation.
+            $appDirs = @(Get-ChildItem -Path $productRoot -Directory -Filter 'app-*' -ErrorAction SilentlyContinue)
+            if ($appDirs.Count -ne 1) {
+              throw "assertion D failed: expected exactly one app-* dir under $productRoot, found $($appDirs.Count): $($appDirs.FullName -join ', ')"
+            }
+
+            Write-Host "=== OK: $exePath ==="
+          }
+
+          Test-WarmLaunchCaching `
+            (Resolve-Path 'RouteConverterWindows\target\RouteConverterWindows.exe').Path `
+            "$env:LOCALAPPDATA\RouteConverter" `
+            'RouteConverterWindows.jar'
+
+          Test-WarmLaunchCaching `
+            (Resolve-Path 'TimeAlbumProWindows\target\TimeAlbumProWindows.exe').Path `
+            "$env:LOCALAPPDATA\TimeAlbumPro" `
+            'TimeAlbumProWindows.jar'
+
       - name: Cache PortableApps.com Installer
         id: pa-cache
         uses: actions/cache@v6
--- a/RouteConverterWindows/src/main/resources/RouteConverter.nsi
+++ b/RouteConverterWindows/src/main/resources/RouteConverter.nsi
@@ -11,6 +11,12 @@ ShowInstDetails hide
 !define JRE_SRC "..\jre-${JRE}"
 ; runtime extraction target (persisted across launches, namespaced under LocalAppData)
 !define JRE_DIR "$LOCALAPPDATA\RouteConverter\jre-${JRE}"
+; app jar generation, namespaced by build so a running instance's directory can
+; never be pulled out from under it; see issue #372.
+!define VERSION "${project.version}-${maven.build.number}"
+!define PRODUCT_DIR "$LOCALAPPDATA\RouteConverter"
+!define APP_DIR "${PRODUCT_DIR}\app-${VERSION}"
+!define STAMP_CONTENT "${VERSION}|${JRE}"
 OutFile "RouteConverterWindows.exe"
 
 Icon "RouteConverter.ico"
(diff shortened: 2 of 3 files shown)


cpesch:

### Panel

  The new stamp-based caching in both .nsi launchers has two runtime bugs that undercut its own stated guarantees: SetOverwrite off silently defeats the crash-safety re-extraction path, and the prune loop can delete unrelated entries under the product directory, not just superseded app-*/jre-* generations. A register-reuse concern is flagged at lower confidence since the rest of the .nsi file is not visible in this diff.

### RouteConverterWindows/src/main/resources/RouteConverter.nsi

- **major** [correctness] line 56 — The prune loop only skips `.`, `..`, the current `app-${VERSION}`, and the current `jre-${JRE}` — every other entry directly under `${PRODUCT_DIR}` (`$LOCALAPPDATA\RouteConverter`) is passed to `RMDir /r`, not just superseded `app-*`/`jre-*` generations. Any other file or folder RouteConverter ever places there (settings, logs, cache, a future config dir) would be silently deleted on the next cold launch, contradicting the comment's stated intent of pruning "superseded generations."  _(correctness)_

```suggestion
Restrict the delete to entries actually matching an `app-*` or `jre-*` naming pattern (e.g. compare a prefix instead of excluding only the two current names) before calling RMDir /r.
```
- **major** [correctness] line 67 — `SetOverwrite off` (unchanged, line 33) is still in effect when `File "RouteConverterWindows.jar"` runs on the slow path, so if a prior launch crashed mid-extraction and left a truncated jar at `${APP_DIR}\RouteConverterWindows.jar` with no valid `.stamp`, this `File` command will NOT overwrite the existing truncated file — it silently skips it because the target already exists. The next launch then runs java against the same truncated jar again, which is exactly the failure mode the new commit-marker comment (lines 69-70) claims is prevented.  _(correctness)_

```suggestion
Add `SetOverwrite on` (or `ifnewer`) immediately before the `SetOutPath "${APP_DIR}"` / `File` pair on the slow path, or delete the stale `${APP_DIR}` directory before extracting when the stamp check fails.
```
- **nit** [correctness] line 38 — The new code uses global registers $1-$4 directly without Push/Pop. NSIS convention reserves $R0-$R9 for user-script scratch use precisely to avoid clobbering state used by other Sections/Functions/plugins in the same script; only this Section's diff is visible here, so it's unclear whether $1-$4 are used elsewhere in the full .nsi file. Flagging at reduced confidence/severity since the surrounding script isn't visible in this diff.  _(correctness)_

```suggestion
Prefer $R0-$R9 (or Push/Pop $1-$4) for the new stamp/prune logic to avoid any risk of clobbering registers used elsewhere in the script.
```

### TimeAlbumProWindows/src/main/resources/TimeAlbumPro.nsi

- **major** [correctness] line 56 — Same scope bug as RouteConverter.nsi: the prune loop's exclusion list only covers `.`, `..`, and the two current version dirs, so any unrelated file/folder under `${PRODUCT_DIR}` gets `RMDir /r`'d rather than only stale `app-*`/`jre-*` generations.  _(correctness)_

```suggestion
Same fix as the RouteConverter.nsi counterpart — match by name pattern, not by exact exclusion of the current version strings.
```
- **major** [correctness] line 67 — Same issue as RouteConverter.nsi: `SetOverwrite off` (line 33) means `File "TimeAlbumProWindows.jar"` on the slow path will not replace an existing truncated jar left by a crashed prior extraction, defeating the crash-safety guarantee described in the adjacent comment.  _(correctness)_

```suggestion
Same fix as the RouteConverter.nsi counterpart — force overwrite (or remove the stale APP_DIR) before re-extracting on the slow path.
```
- **nit** [correctness] line 38 — Same register-reuse concern as RouteConverter.nsi: $1-$4 are used directly rather than the conventional $R0-$R9 scratch registers, with the rest of the script not visible in this diff to confirm there's no collision.  _(correctness)_

```suggestion
Same as RouteConverter.nsi — consider $R0-$R9 for consistency and safety.
```

---

PR #377, opened 2026-09-03
world map background leaves a permanent grey hole after startup

### LINKED ISSUE #376: [map] world map background leaves a permanent grey hole after startup
## Bug: world-map background leaves a permanent grey hole after startup

**Source:** maintainer session 2026-09-03, while verifying the server-side `world.map`
replacement (`static.routeconverter.com/maps/world.map` now carries `natural=sea` /
`natural=nosea` polygons instead of bare `natural=coastline`). Reproduced independently
by the maintainer and by a local build of `master` — same rectangle, same place.

Symptom: the world background renders, but a large axis-aligned rectangle stays flat
grey for the whole session. It does not repaint on its own.

### What is established

**The grey is `#EEEEEE` = `DisplayModel.DEFAULT_BACKGROUND_COLOR`**
(mapsforge `mapsforge-map/.../model/DisplayModel.java:30`). That is the frame-buffer
clear colour, so the region is *unpainted by any layer* — not a wrong fill.

**It is not the map data.** Rendering all 20 zoom-4 tiles covering Europe
(x 6..10, y 3..6) straight out of the live `world.map` through `DatabaseRenderer`,
with `hasAlpha=true` (matching how the background layer renders) and with the
maintainer's applied theme `openandromaps/Lightning.xml`, every tile comes back
`paintedPixels=65536 / 65536 (100.0%)`. `store.supportsTile(tile)` is true for all of them.

**The hole is exactly two tiles.** The background layer's `FileSystemTileCache`
(`$TMPDIR/routeconverter-$USER/world.map/<z>/<x>/<y>.tile`) held 18 of those 20
tiles after the run — missing precisely `4/8/5` and `4/8/6`. Tile x=8 at z4 spans
0°..22.5°E and y=5,6 span ~56°N down to ~22°N, which is the rectangle on screen.

**A tile render was interrupted one second after the background map attached:**

```
09:00:33 MapsforgeMapView setBackgroundMap
  INFORMATION: Loaded background map .../world.map (3276950 bytes)
09:00:34 BaseRouteConverter adjustDividersForScreenMovement
  INFORMATION: Adjusted profile divider to 791
09:00:34 org.mapsforge.map.layer.renderer.DatabaseRenderer executeJob
  WARNUNG: java.lang.InterruptedException
09:00:34 MapsforgeMapView handleMapAndThemeUpdate
  INFORMATION: Using map [MapsforgeFileMap[...schleswig-holstein.map]] and theme
               VectorTheme[...Lightning.xml] with zoom 4
```

`DatabaseRenderer.executeJob` catches the interrupt, logs it, and returns `null`.
The tile is then neither cached nor drawn.

**Why nothing repairs it.** `TileLayer.draw()` does re-queue a missing tile
(`if (bitmap == null) { if (hasJobQueue && !tileCache.containsKey(job)) jobQueue.add(job); }`),
so the hole is recoverable — but only on the *next* draw. After startup settles nothing
repaints the map view, so the gap survives until the user pans or zooms.

**Why it is visible at all.** [`DefaultTileLayerFactory:86`](../blob/master/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/tiles/DefaultTileLayerFactory.java#L86)
builds every `TileRendererLayer` — displayed map *and* background — with
`isTransparent=true`, so a tile that never renders leaves bare canvas rather than
any background fill. This was always broken; it only became visible now that the
world map has something to show behind the hole. (Note for whoever fixes this:
flipping the background layer to `isTransparent=false` is **not** the fix —
`TileLayer.draw()` then calls `canvas.fillColor(displayModel.getBackgroundColor())`,
i.e. it paints the same `#EEEEEE`, not the theme background.)

### Still to confirm before implementing

One check the fix depends on, not yet done: **does panning or zooming fill the hole
in?** If yes, the diagnosis above is complete and the fix is "force one redraw after
the startup layer-stack rebuild". If the hole survives a pan/zoom, the interrupted job
is being suppressed by something else (e.g. `TwoLevelTileCache.containsKey` returning
true for a second-level-only entry while `getImmediately` returns null, which would
permanently stop the re-queue) and this issue needs re-scoping first.

Only one `InterruptedException` was logged against two missing tiles, so at least one
tile was lost without a log line — worth accounting for.

### Proposed fix (pending the check above)

`MapsforgeMapView.handleMapAndThemeUpdate()` tears down and rebuilds the layer stack
(`layers.remove(...)`, `remove.onDestroy()`, `tileLayer.getTileCache().destroy()`)
and then re-attaches the background via `handleBackground()`. When the async download
callback path (`setBackgroundMap`) wins the race and attaches the background *first*,
its in-flight tile jobs are killed by that rebuild and never re-issued.

After the rebuild completes and a background layer is attached, request one redraw of
the layer stack so any job dropped during the rebuild is re-queued.

### Files

- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java`
  (`handleMapAndThemeUpdate`, `handleBackground`, `setBackgroundMap`)
- `mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachment.java`
- `mapsforge-mapview/src/test/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachmentTest.java`

`BackgroundMapAttachment` already exists to make exactly this Path A / Path B ordering
race testable as a pure decision, with `BackgroundMapAttachmentTest` as the pattern.
Add a sibling predicate there rather than burying the condition in the view:

```java
public static boolean shouldRedrawAfterStackRebuild(boolean backgroundAttached, boolean stackRebuilt)
```

Cases to cover:
- `(false, true)` -> false — nothing attached, nothing to repair
- `(true, false)` -> false — no rebuild happened, do not force redraws on every update
- `(true, true)`  -> true  — background attached and the stack was rebuilt under it

### Acceptance criteria

- `BackgroundMapAttachmentTest` covers the three cases above; `mvn -pl mapsforge-mapview test` green.
- Manual: start with a partial offline map selected (e.g. `kurviger/.../schleswig-holstein.map`)
  and the world background installed; the whole viewport shows land/sea immediately at
  startup, with no flat `#EEEEEE` rectangle, without touching the map.
- No `DatabaseRenderer executeJob ... InterruptedException` left unrepaired: if one is
  logged, the affected tile is painted by the time the view settles.

### Reproduction / diagnostic harness

Headless, no GUI needed — renders tiles straight from a `.map` file through
`DatabaseRenderer` and reports painted-pixel coverage per tile, which is what separates
"map data is wrong" from "the client did not paint it":

```java
DatabaseRenderer r = new DatabaseRenderer(store, AwtGraphicFactory.INSTANCE,
        new InMemoryTileCache(64), null, true, true, null);
TileBitmap tb = r.executeJob(new RendererJob(tile, store, renderThemeFuture, displayModel, 1f, transparent, false));
```

Gotcha: pass a non-null `TileCache`. With `null` the renderer throws
`NullPointerException: ... "this.tileCache" is null` inside `executeJob`, which is
swallowed into a warning and returns `null` for every tile after the first.

cpesch:
## Summary

Fixed the permanent grey hole in the world-map background (issue #376) by forcing one layer-stack redraw after `handleMapAndThemeUpdate()` rebuilds the map/theme layers, so a background tile job interrupted by that rebuild gets re-queued by mapsforge's `TileLayer.draw()` instead of staying unpainted for the rest of the session.

**Root cause (per the issue's diagnosis):** `handleMapAndThemeUpdate()` tears down and rebuilds the displayed-map layer stack on every map/theme change. If the async background-download callback (`setBackgroundMap`) had already attached the background layer and had tile jobs in flight, that rebuild kills them without re-issuing them. Nothing repaints the view afterward, so the hole (interrupted tiles) persists until the user pans or zooms.

**Changes:**
- `BackgroundMapAttachment.java` — added the sibling predicate `shouldRedrawAfterStackRebuild(boolean backgroundAttached, boolean stackRebuilt)`, covering exactly the three cases from the issue: no background attached → false; background attached but no rebuild happened → false (don't force redraws on every update); both true → force a redraw.
- `BackgroundMapAttachmentTest.java` — added the three corresponding unit tests.
- `MapsforgeMapView.java` — split `handleBackground()` into a no-arg wrapper (`stackRebuilt=false`, used by `setBackgroundMap`) and `handleBackground(boolean stackRebuilt)`. `handleMapAndThemeUpdate()` now calls `handleBackground(true)`; when the predicate is true, it calls `getLayerManager().redrawLayers()` (the same mechanism already used elsewhere in this file to force a repaint), which lets `TileLayer.draw()` re-queue the tile job the rebuild dropped.

**Verification:** unavailable in this environment — `factory-verify` returned a broker error (`repo not allowed (rc/* only)`), and no JDK is reachable on `PATH` or via sdkman outside the sandboxed working directory, so I could not run `mvn -pl mapsforge-mapview test` myself. The new tests are plain JUnit against the pure predicate (mirroring the existing `attachesWhenLayerReadyAndAnyMapDisplayed` pattern) and should pass; please run `./mvnw -pl mapsforge-mapview -am test` to confirm, and do the manual startup check described in the issue (partial offline map + world background, no flat `#EEEEEE` rectangle after startup).

Source issue: cpesch/RouteConverter#376

---
**Builder stats** · model `sonnet` · confidence `0` · 2m 2s across 1 round(s) · 3 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/26215)

Closes #376

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachment.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachment.java
@@ -47,4 +47,18 @@ public class BackgroundMapAttachment {
     public static boolean shouldAttachBackground(boolean layerReady, boolean hasDisplayedMap) {
         return layerReady && hasDisplayedMap;
     }
+
+    /**
+     * A layer-stack rebuild (map/theme change) can kill in-flight tile jobs of an already
+     * attached background layer without re-issuing them, leaving unpainted tiles behind
+     * until the user pans or zooms. Forcing one redraw afterwards lets
+     * {@code TileLayer.draw()} re-queue whatever job the rebuild dropped.
+     *
+     * @param backgroundAttached    whether the background layer is attached as the base layer now
+     * @param stackRebuilt          whether the displayed map/theme layer stack was just torn down and rebuilt
+     * @return true iff one redraw of the layer stack should be forced to repair a job dropped during the rebuild
+     */
+    public static boolean shouldRedrawAfterStackRebuild(boolean backgroundAttached, boolean stackRebuilt) {
+        return backgroundAttached && stackRebuilt;
+    }
 }
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/MapsforgeMapView.java
@@ -751,7 +751,7 @@ public class MapsforgeMapView extends BaseMapView {
         layers.add(0, layer);
         mapsToLayers.put(map, layer);
 
-        handleBackground();
+        handleBackground(true);
         handleOverlays();
         handleTrackLayer();
         handleNonSelectedPositionLists();
@@ -793,13 +793,25 @@ public class MapsforgeMapView extends BaseMapView {
     }
 
     private void handleBackground() {
+        handleBackground(false);
+    }
+
+    // stackRebuilt is true when called from handleMapAndThemeUpdate(), which tears down and
+    // rebuilds the displayed map layer stack; an in-flight tile job of an already attached
+    // background layer can be killed by that rebuild and never re-issued (issue #376), so
+    // force one redraw afterwards to let TileLayer.draw() re-queue the dropped job
+    private void handleBackground(boolean stackRebuilt) {
         Layers layers = getLayerManager().getLayers();
         if (backgroundLayer != null)
             layers.remove(backgroundLayer);
 
         LocalMap map = getMapManager().getDisplayedMapModel().getItem();
-        if (BackgroundMapAttachment.shouldAttachBackground(backgroundLayer != null, map != null))
+        boolean backgroundAttached = BackgroundMapAttachment.shouldAttachBackground(backgroundLayer != null, map != null);
+        if (backgroundAttached)
             layers.add(0, backgroundLayer);
+
+        if (BackgroundMapAttachment.shouldRedrawAfterStackRebuild(backgroundAttached, stackRebuilt))
+            getLayerManager().redrawLayers();
     }
 
     private void handleNonSelectedPositionLists() {
--- a/mapsforge-mapview/src/test/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachmentTest.java
+++ b/mapsforge-mapview/src/test/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachmentTest.java
@@ -24,6 +24,7 @@ import org.junit.Test;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static slash.navigation.mapview.mapsforge.BackgroundMapAttachment.shouldAttachBackground;
+import static slash.navigation.mapview.mapsforge.BackgroundMapAttachment.shouldRedrawAfterStackRebuild;
 
 /**
  * Tests for {@link BackgroundMapAttachment}.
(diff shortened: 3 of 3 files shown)


cpesch (commented):

### Panel

- **correctness** — approve · confidence 0.87
  Predicate and wiring match the issue's diagnosis: the rebuild path (handleMapAndThemeUpdate -> handleBackground(true)) forces exactly one extra redrawLayers() only when a background layer is attached, and the no-arg wrapper preserves old behaviour for any caller outside the diff, so the worst case is a coalesced redundant repaint. No runtime harm found; note the fix's efficacy rests on the unverified claim that TileLayer.draw() re-queues dropped jobs, and no test in this PR was actually executed (no JDK in the build environment) — please run the mapsforge-mapview tests plus the manual startup check before merging.

---

PR #379, opened 2026-09-04
ci(crowdin): pin manual disable in-file — if:false + banner (rc-meta#127)

cpesch:
Makes the manual UI-disable of the Crowdin sync workflow self-documenting in the file itself: `if: false` on the single job plus a banner explaining why it is off (rc/rc-meta#127), what secrets it consumes, and that re-enabling requires removing the `if: false` AND re-enabling in the UI together. The factory's GH_BUILDER_PAT deliberately lacks the workflow scope, so this file only ever changes via a maintainer-credentialed push.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -1,3 +1,17 @@
+# ---------------------------------------------------------------------------
+# DISABLED MANUALLY (see rc/rc-meta#127) — do not re-enable without review.
+#
+# This workflow pushes translation commits and opens pull requests using
+# repo secrets (GITHUB_TOKEN, CROWDIN_PROJECT_ID, CROWDIN_PERSONAL_TOKEN).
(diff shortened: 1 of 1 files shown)


cpesch (commented):

### Panel

- **correctness** — approve · confidence 0.92
  Annotation-only change: a comment banner plus a job-level `if: false` on the single `synchronize` job. `if: false` is the standard idiom for disabling a job and fails safe — the job reports skipped on every trigger (schedule or dispatch), so nothing resumes unless the flag is removed, and removing the UI-disable without removing the flag also stays skipped. Banner claims about token scope align with the recorded AGENTS.md facts rather than contradicting them.

---

cpesch:

**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-26439` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-26439).

PR #380, opened 2026-09-05
ci(crowdin): gate on CROWDIN_SYNC_ENABLED var — if:false trips actionlint

cpesch:
Follow-up to #379: actionlint's if-cond rule rejects the constant `if: false`, turning Lint workflows red on every push since the merge. Same disabled-by-default semantics via the repository variable `CROWDIN_SYNC_ENABLED` (unset today); banner updated — re-enable = set the var AND re-enable in the UI together.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -4,9 +4,12 @@
 # This workflow pushes translation commits and opens pull requests using
 # repo secrets (GITHUB_TOKEN, CROWDIN_PROJECT_ID, CROWDIN_PERSONAL_TOKEN).
 # Re-enabling requires a maintainer to review current CROWDIN_* secret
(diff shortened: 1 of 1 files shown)


cpesch (commented):
## Verdict: comment-only · confidence 0.93

### Panel

- **correctness** — approve · confidence 0.93
  Job-level gate swaps the constant `if: false` for `vars.CROWDIN_SYNC_ENABLED == 'true'`: an unset repository variable resolves to an empty string, so the comparison is false and the job stays skipped — disabled-by-default semantics are preserved, and any value other than the exact string `true` (e.g. `True`, whitespace, or a secret created instead of a variable) also fails closed toward disabled. `vars` is a valid context in a job-level `if`, so no runtime regression is visible from this lens.

---

PR #381, opened 2026-09-06
fix(map): keep the world-map background rendering after the layer stack is rebuilt

cpesch:
## Summary

The world-map background stopped rendering right after startup and stayed blank for the whole session (#376, and the "grey hole" that #377 chased). Both earlier fixes addressed symptoms; the layer itself was dead.

## Root cause

mapsforge 0.30.0 bug (still on mapsforge master, reported upstream as mapsforge/mapsforge#1817): `Layers.add()` calls `setDisplayModel()` on every add, `TileLayer.setDisplayModel()` replaces the layer's `JobQueue` with a fresh instance each time, but `TileRendererLayer.setDisplayModel()` creates its `MapWorkerPool` only once, bound to the first queue. After a remove + re-add, `TileLayer.draw()` queues tile jobs into the new queue while the worker pool waits on the old one forever.

`handleBackground()` always removed and re-added the background layer to move it below the freshly inserted displayed-map layer. So the first `handleMapAndThemeUpdate()` after `setBackgroundMap()` killed the background for good. The forced redraw from #377 only fed the orphaned queue.

Evidence: the background's `FileSystemTileCache` dir (`$TMPDIR/routeconverter-$USER/world.map/`) stayed empty while the displayed map cached 58 tiles; a thread dump showed the background's `MapWorkerPool` loop parked in `JobQueue.get()` with its worker threads never spawned; the heap held two `JobQueue`s for that layer.

## Changes

- `ReattachableTileRendererLayer` keeps the first `JobQueue` for the lifetime of the layer so `draw()` and the worker pool stay on the same queue. `DefaultTileLayerFactory` creates it for displayed maps and the background.
- `handleMapAndThemeUpdate()` inserts the displayed map directly above an attached background (`BackgroundMapAttachment.displayedMapLayerIndex`) instead of at index 0, so the background never has to move. `handleBackground()` only attaches or detaches on a state change.
- `setBackgroundMap()` detaches and destroys the previous background layer when `world.map` is re-downloaded mid-session instead of leaving it in the layer list.
- Dropped the unused no-arg `handleBackground()`.

## Verification

- Headless harness: a plain `TileRendererLayer` renders 0 tiles after remove + re-add, the reattachable one renders all 20.
- `ReattachableTileRendererLayerTest` (2 tests, one documents the upstream bug so a mapsforge upgrade that fixes it makes the workaround removable), `BackgroundMapAttachmentTest` +2. `mvn -pl mapsforge-mapview test`: 153 tests green.
- Running app with `schleswig-holstein.map` + world background: `world.map` tile cache fills at startup (59 tiles), whole viewport shows land and sea at zoom 5 with no grey area, no `InterruptedException` in the log.

Refs #376, supersedes #377.

### DIFF
--- a/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachment.java
+++ b/mapsforge-mapview/src/main/java/slash/navigation/mapview/mapsforge/BackgroundMapAttachment.java
@@ -49,6 +49,19 @@ public class BackgroundMapAttachment {
     }
 
     /**
+     * The displayed map layer has to sit directly above the world-map background. Inserting it there
+     * instead of always at index 0 and then moving the background back below it avoids removing and
+     * re-adding the background layer, which mapsforge's {@code TileRendererLayer} does not survive
+     * (see {@code ReattachableTileRendererLayer}).
+     *
+     * @param backgroundLayerIndex  index of the attached background layer in the layer list, or -1 if none is attached
+     * @return the index at which the displayed map layer has to be inserted
+     */
+    public static int displayedMapLayerIndex(int backgroundLayerIndex) {
+        return backgroundLayerIndex < 0 ? 0 : backgroundLayerIndex + 1;
+    }
+
+    /**
      * A layer-stack rebuild (map/theme change) can kill in-flight tile jobs of an already
      * attached background layer without re-issuing them, leaving unpainted tiles behind
      * until the user pans or zooms. Forcing one redraw afterwards lets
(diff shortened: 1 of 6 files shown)


cpesch (commented):

### Panel

- **correctness** — approve · confidence 0.85
  The queue-reuse fix and the attach/detach state tracking in handleBackground/handleMapAndThemeUpdate are internally consistent and covered by targeted tests (including a test that documents the upstream mapsforge bug); no runtime-correctness defects found in the diff's changed lines.

---

PR #385, opened 2026-09-06
ChecksumSender reports the checksum of a download that failed validation, polluting the catalog

### LINKED ISSUE #382: [download] ChecksumSender reports the checksum of a download that failed validation, polluting the catalog
## Bug: `ChecksumSender.failed()` publishes the checksum of a download that just failed validation

**Source:** maintainer session 2026-09-06, while fixing the world-map background (#376, #381). Server side: `routeconverter-maps/world.map` had collected 8 bogus `MapChecksum` rows between 2026-09-04 and 2026-09-06 (contentLength 1662900 / 1662901 / 1662902, 2535723, 2774898 / 2774899 / 2774900, 2358898 against the real 3276950-byte file), 7 of them anonymous. The same thing happened 2026-08-21..30 (9 rows) and 2026-06-28. Every time, the newest bogus rows pushed the real checksum out of the served top-5 and **every client re-downloaded world.map on each start**, logging `world.map has SHA-1 … but matched none of 5 known-good checksum(s)`.

The server now quarantines anonymous reports until a second client corroborates them (rc-site #52, deployed 2026-09-06), so the catalog no longer breaks. The client still produces the garbage; this issue is the client half.

### What is established

The pollution path is in the download module plus `ChecksumSender`:

1. `GetPerformer.run()` (download/…/performer/GetPerformer.java:133): the HTTP transfer completes (`result.success`), `postProcess()` calls `bringToTarget()` and then `validate()`.
2. `validate()` (:221) runs `Validator.isChecksumsValid()`, which computes SHA-1 / size of the **file just written** and compares against the catalog. Mismatch → `ChecksumError` → `downloadExecutor.postProcessFailed()` → `DownloadManager.fireFailed()`.
3. `ChecksumSender.failed()` (route-converter-gui/…/helpers/ChecksumSender.java:41):
   ```java
   if (file.getActualChecksum().laterThan(file.getExpectedChecksum()) ||
           file.getExpectedChecksum().getSHA1() == null && file.getActualChecksum().getSHA1() != null)
       sendChecksums(download);
   ```
   `Checksum.laterThan()` compares `lastModified` only. A freshly written file is practically always "later" than the catalog row, so **the checksum of the content that failed validation is sent to the server as a known-good checksum** (`BaseRouteConverter.sendChecksums` → `RouteFeedback.sendChecksums`, unauthenticated PUT).

The branch exists for a real case: the upstream rebuilt a file (BRouter tiles, #155) and the catalog is stale, so the client tells the server about the new build. It cannot tell that case apart from a truncated or otherwise wrong download. The observed sizes (three rows differing by exactly one byte on consecutive days, twice) are not builds; they look like interrupted transfers, possibly through `resume()` (:77) appending to an existing temp file with a `Range` request.

Also `downloadFailed()` (I/O exception path) fires the same `failed()` listener; whatever `actualChecksum` happens to be set (from `updateDownload(…, false)` or an earlier validation) is reported the same way.

### Proposed fix

Report a failed download's checksum only when the transfer is provably complete, otherwise stay silent:

- the HTTP response announced a `Content-Length` (`ReadRequest.getContentLength()`) and the written file has exactly that size, and
- the server sent a `Last-Modified` and it equals the file's `lastModified` (this is what `laterThan` was approximating), and
- the failure state is `ChecksumError` (validation ran on a complete file), not a transport failure.

A `ChecksumError` on a complete, correctly sized file is the genuine "upstream rebuilt" case and may be reported. A size mismatch against the announced `Content-Length`, a missing `Content-Length`, or a transport failure must not be. Keep `succeeded()` as is.

`Download` does not currently retain the announced `Content-Length` after `updateDownload()`; `extractChecksum(request)` folds it into `actualChecksum` only when the target file does not exist yet. Carry the announced length (and `Last-Modified`) on the `Download` so the listener can compare.

### Files

- `route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java` (`failed`)
- `download/src/main/java/slash/navigation/download/performer/GetPerformer.java` (`updateDownload`, `validate`, `postProcess`)
- `download/src/main/java/slash/navigation/download/Download.java` (carry announced Content-Length / Last-Modified)
- new pure helper + test, e.g. `download/src/main/java/slash/navigation/download/ChecksumReportPolicy.java` and `download/src/test/java/slash/navigation/download/ChecksumReportPolicyTest.java`

### Testable seam

```java
// true iff the checksum of a download that failed validation may be reported to the server as a new known-good build
public static boolean shouldReportFailedDownload(State state, Long announcedContentLength, Long actualContentLength,
                                                 Long announcedLastModified, Long actualLastModified)
```

Cases to cover:
- `ChecksumError`, announced 3276950, actual 3276950, same lastModified → **true** (genuine rebuild)
- `ChecksumError`, announced 3276950, actual 1662901 → **false** (truncated)
- `ChecksumError`, announced `null` → **false** (cannot prove completeness)
- `ChecksumError`, sizes equal, lastModified differs → **false**
- `Failed` (transport error), sizes equal → **false**
- `Succeeded` is not routed through this predicate (unchanged behaviour)

### Acceptance criteria

- `ChecksumReportPolicyTest` covers the cases above; `mvn -pl download,route-converter-gui test` green.
- Manual: replace `~/.routeconverter/maps/routeconverter/world.map` with a truncated copy (`head -c 1662901`), start RouteConverter. The log shows the `matched none of … known-good checksum(s)` warning and the re-download, but **no** `Sending checksums for {FileAndChecksum[file=…world.map, … actualChecksum=Checksum[… contentLength=1662901 …]]}` line for the truncated content. After the re-download completes, the checksum of the complete file is sent as before (`succeeded()` path).
- No regression for #155: a file whose catalog entry is stale but whose download is complete (sizes equal, same Last-Modified) still gets reported.

### Related

- server side: rc-site #52 (quarantine anonymous reports, cleanup keeps verified rows), deployed 2026-09-06
- #155 (accept any known-good build), #190 (rd5 HEAD-only ingest)
- #376 / #381 (world-map background; the symptom that led here)

cpesch:
Build tooling (`mvnw`, `mvn`, `factory-verify`) is blocked in this session, and `factory-verify`'s broker additionally rejects this repo (`repo not allowed (rc/* only)`), so I could not compile or run tests. Here is the summary.

## Verification result: **unavailable**

- `factory-verify` fails with a broker error: `repo not allowed (rc/* only)` — this checkout is not an allowlisted repo.
- `./mvnw` / `mvn` invocations are denied by the tool policy; no JDK is on PATH (`java: command not found`).
- In lieu of compiling, I traced every API I touched against its source: `ReadRequest.getContentLength()`/`getLastModified()`, `CompactCalendar.getTimeInMillis()`, `Transfer.roundMillisecondsToSecondPrecision`, the `Result` record constructor chain in `GetPerformer`, and the existing `GetPerformerTest` harness (`/ok` serves a 200 with `Content-Length`, `/missing` serves 404). All line up.

## What changed and why

`ChecksumSender.failed()` used `Checksum.laterThan()` — a last-modified comparison that a freshly written file almost always wins — to decide whether to publish a failed download's checksum. Any truncated or otherwise broken transfer therefore got reported as a known-good build, which repeatedly evicted the real `world.map` checksum from the server's served top-5 and forced a global re-download. The fix reports a failed download's checksum **only when the transfer provably completed**:

- **`download/…/ChecksumReportPolicy.java` (new)** — pure predicate `shouldReportFailedDownload(state, announcedContentLength, actualContentLength, announcedLastModified, actualLastModified)` as specified in the issue. Requires `ChecksumError` (validation ran on a complete file), an announced Content-Length exactly equal to the written size, and an announced Last-Modified equal to the file's mtime (both sides rounded to second precision, since HTTP dates and filesystems carry no milliseconds). Missing/`null` inputs fail closed.
- **`download/…/Download.java`** — carries `announcedContentLength` / `announcedLastModified` (the response's `Content-Length` / `Last-Modified`), which were previously folded into `actualChecksum` only when the target file didn't yet exist, so the listener could no longer see them.
- **`download/…/performer/GetPerformer.java`** — clears both announced values at the start of `run()` so a previous attempt's values can't vouch for this attempt; sets them on the success path from the response headers via an extended `Result` record. The length is captured in `download()`'s handler rather than read back through `request` in `run()`: on the `resume()` path a 206's `Content-Length` describes only the remaining range, so a value read later would misrepresent the resource — the full-body `download()` path is the only place a whole-resource length exists.
- **`route-converter-gui/…/helpers/ChecksumSender.java`** — `failed()` now delegates to the policy. `succeeded()` is unchanged; the `#155` case (stale catalog entry, complete download) still reports.
- **`download/src/test/…/ChecksumReportPolicyTest.java` (new)** — 12 tests covering every case in the issue's seam list (genuine rebuild → true; truncated, oversize, missing lengths, last-modified mismatch, transport failure, `NoFileError`, `Succeeded` → false) plus sub-second mtime drift tolerance.
- **`download/src/test/…/performer/GetPerformerTest.java`** — two hermetic wiring tests: a `ChecksumError` download retains the announced `Content-Length` matching the served body, and a `Failed` transfer clears both announced values.

The manually-verifiable acceptance path stands: with a truncated `world.map`, the log should show the `matched none of … known-good checksum(s)` warning and the re-download, but no `Sending checksums … contentLength=1662901` line; the complete re-download still reports via `succeeded()`.

Source issue: cpesch/RouteConverter#382

---
**Builder stats** · model `sonnet` · confidence `0` · 27m 8s across 1 round(s) · 6 file op(s)
[Workflow run](http://127.0.0.1:3000/cpesch/RouteConverter/actions/runs/27155)

Closes #382

### DIFF
--- a/route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java
+++ b/route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java
@@ -20,6 +20,8 @@
 package slash.navigation.converter.gui.helpers;
 
 import slash.navigation.converter.gui.BaseRouteConverter;
+import slash.navigation.download.Checksum;
+import slash.navigation.download.ChecksumReportPolicy;
 import slash.navigation.download.Download;
 import slash.navigation.download.DownloadListener;
 import slash.navigation.download.FileAndChecksum;
@@ -43,9 +45,17 @@ public class ChecksumSender implements DownloadListener {
         FileAndChecksum file = download.getFile();
         if (file.getActualChecksum() == null)
             return;
-        if (file.getActualChecksum().laterThan(file.getExpectedChecksum()) ||
-                file.getExpectedChecksum().getSHA1() == null && file.getActualChecksum().getSHA1() != null)
+        // the file's checksum is only a new known-good build if the transfer provably completed
+        // and validation failed on the complete content; anything else is a broken download whose
+        // checksum must not be published (see GitHub #382)
+        if (ChecksumReportPolicy.shouldReportFailedDownload(download.getState(),
+                download.getAnnouncedContentLength(), file.getActualChecksum().getContentLength(),
+                download.getAnnouncedLastModified(), lastModifiedMillis(file.getActualChecksum())))
             sendChecksums(download);
+    }
+
+    private Long lastModifiedMillis(Checksum checksum) {
+        return checksum.getLastModified() != null ? checksum.getLastModified().getTimeInMillis() : null;
     }
 
     public void succeeded(Download download) {
--- a/route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java
+++ b/route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java
@@ -48,7 +48,7 @@ public class ChecksumSender implements DownloadListener {
         // the file's checksum is only a new known-good build if the transfer provably completed
         // and validation failed on the complete content; anything else is a broken download whose
         // checksum must not be published (see GitHub #382)
-        if (ChecksumReportPolicy.shouldReportFailedDownload(download.getState(),
+        if (ChecksumReportPolicy.isReportableChecksum(download.getState(),
                 download.getAnnouncedContentLength(), file.getActualChecksum().getContentLength(),
                 download.getAnnouncedLastModified(), lastModifiedMillis(file.getActualChecksum())))
             sendChecksums(download);
--- a/download/src/main/java/slash/navigation/download/ChecksumReportPolicy.java
+++ b/download/src/main/java/slash/navigation/download/ChecksumReportPolicy.java
@@ -41,15 +41,16 @@ public class ChecksumReportPolicy {
     private ChecksumReportPolicy() {}
 
     /**
-     * @param state the terminal state of the failed download
+     * @param state the terminal state of the download
      * @param announcedContentLength the Content-Length announced by the HTTP response, if any
      * @param actualContentLength the size of the file that was written
      * @param announcedLastModified the Last-Modified announced by the HTTP response, if any
      * @param actualLastModified the last-modified of the file that was written
-     * @return true iff the checksum of this download may be reported to the server as a new known-good build
+     * @return true iff this checksum is trustworthy enough to be reported to the server as a
+     * new known-good build
      */
-    public static boolean shouldReportFailedDownload(State state, Long announcedContentLength, Long actualContentLength,
-                                                     Long announcedLastModified, Long actualLastModified) {
+    public static boolean isReportableChecksum(State state, Long announcedContentLength, Long actualContentLength,
+                                               Long announcedLastModified, Long actualLastModified) {
         // only a validation failure happens on a completely transferred file; a transport error
         // leaves the content in an unknown state whose checksum is meaningless
         if (state != State.ChecksumError)
--- a/download/src/main/java/slash/navigation/download/ChecksumReportPolicy.java
+++ b/download/src/main/java/slash/navigation/download/ChecksumReportPolicy.java
@@ -1,0 +1,71 @@
+/*
+    This file is part of RouteConverter.
+
+    RouteConverter is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    RouteConverter is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with RouteConverter; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+    Copyright (C) 2007 Christian Pesch. All Rights Reserved.
+*/
+
+package slash.navigation.download;
+
+import static slash.common.io.Transfer.roundMillisecondsToSecondPrecision;
+
+/**
+ * Decides whether the checksum of a {@link Download} that failed may be reported to the catalog
+ * server as a new known-good build.
+ * <p>
+ * A failed download only carries a reportable checksum if the transfer provably completed and the
+ * failure came from validation: the response announced a Content-Length the written file matches
+ * exactly, and a Last-Modified the file still carries. That is the genuine "the upstream rebuilt
+ * the file and the catalog is stale" case the catalog wants to learn about. Anything else — a
+ * truncated transfer, a response without a Content-Length, a transport error, a different
+ * Last-Modified — describes content the server never announced, and reporting its checksum would
+ * pollute the catalog's known-good checksums (see GitHub #382).
+ *
+ * @author Christian Pesch
+ */
+public class ChecksumReportPolicy {
+
+    private ChecksumReportPolicy() {}
+
+    /**
+     * @param state the terminal state of the failed download
+     * @param announcedContentLength the Content-Length announced by the HTTP response, if any
+     * @param actualContentLength the size of the file that was written
+     * @param announcedLastModified the Last-Modified announced by the HTTP response, if any
+     * @param actualLastModified the last-modified of the file that was written
+     * @return true iff the checksum of this download may be reported to the server as a new known-good build
+     */
+    public static boolean shouldReportFailedDownload(State state, Long announcedContentLength, Long actualContentLength,
+                                                     Long announcedLastModified, Long actualLastModified) {
+        // only a validation failure happens on a completely transferred file; a transport error
+        // leaves the content in an unknown state whose checksum is meaningless
+        if (state != State.ChecksumError)
+            return false;
+
+        // without an announced Content-Length the transfer cannot be proven complete, and a size
+        // that differs from it means the content is not the announced one
+        if (announcedContentLength == null || actualContentLength == null ||
+                !announcedContentLength.equals(actualContentLength))
+            return false;
+
+        // without an announced Last-Modified the file cannot be proven to be the announced build;
+        // both sides are truncated to second precision, as HTTP dates and file systems carry no
+        // milliseconds
+        return announcedLastModified != null && actualLastModified != null &&
+                roundMillisecondsToSecondPrecision(announcedLastModified) ==
+                        roundMillisecondsToSecondPrecision(actualLastModified);
+    }
+}
--- a/download/src/main/java/slash/navigation/download/Download.java
+++ b/download/src/main/java/slash/navigation/download/Download.java
@@ -53,6 +53,8 @@ public class Download {
     private State state;
     private long processedBytes;
     private Long expectedBytes;
+    private Long announcedContentLength;
+    private Long announcedLastModified;
 
     public Download(String description, String url, Action action, FileAndChecksum file,
                     List<FileAndChecksum> fragments, String eTag, State state, File tempFile) {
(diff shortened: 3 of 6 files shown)


cpesch:

### Panel

  The predicate itself is fail-closed and its state gate is sound, but both load-bearing proofs rest on unchanged code outside this diff — whether GetPerformer.resume() still calls the removed Result overload, and whether the announced Last-Modified actually reaches the written file's checksum — while the PR description records that the module was never compiled or its tests run. Resume-completed downloads carry no announced values and therefore fail closed by design; the supplied diff slice is complete, so no file-omission caveat applies.

### download/src/main/java/slash/navigation/download/ChecksumReportPolicy.java

- **minor** [correctness] line 65 — Using == between two roundMillisecondsToSecondPrecision results is reference comparison if that helper returns boxed Long, and at epoch-second magnitudes (far outside the Long autobox cache) it would then never match, disabling the last-modified leg; if the helper returns primitive long this is fine.  _(correctness)_

```suggestion
Bind both results to long locals or compare with .equals() so the check is correct regardless of the helper's return type.
```

### download/src/main/java/slash/navigation/download/performer/GetPerformer.java

- **major** [correctness] line 258 — The diff removes the Result(Get, boolean, Long lastModified) overload, whose only caller visible in the diff (download()'s success return) was migrated — but resume(), the only other Result-producing site in GetPerformer and outside the shown hunks, may still call the removed overload, and the PR description states the module was never compiled, so a leftover caller would go uncaught.  _(correctness)_

```suggestion
Check every remaining `new Result(` in GetPerformer (or run `mvn -pl download compile` once a JDK is available) and migrate any `(request, success, lastModified)` call to the new 4-arg form before merge.
```

### route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java

- **minor** [correctness] line 53 — The last-modified leg compares the announced Last-Modified against actualChecksum.getLastModified(), whose provenance is not established in this diff: if that value is the local write time (as the old laterThan behaviour of a freshly written file winning the comparison implies) and nothing sets the file/checksum mtime from the response header, the predicate can never return true and the genuine-rebuild report path is dead at runtime — fail-closed, but the stale-catalog case this policy exists to enable silently stops reporting.  _(correctness)_

```suggestion
Pin the chain with a hermetic test asserting that for a completed download download.getAnnouncedLastModified() equals the served Last-Modified and equals checksum.getLastModified().getTimeInMillis().
```

---

cpesch:

**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-27163` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-27163).

route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java
```diff
@@ -43,11 +45,19 @@ public void failed(Download download) {
         FileAndChecksum file = download.getFile();
         if (file.getActualChecksum() == null)
             return;
-        if (file.getActualChecksum().laterThan(file.getExpectedChecksum()) ||
-                file.getExpectedChecksum().getSHA1() == null && file.getActualChecksum().getSHA1() != null)
+        // the file's checksum is only a new known-good build if the transfer provably completed
+        // and validation failed on the complete content; anything else is a broken download whose
+        // checksum must not be published (see GitHub #382)
+        if (ChecksumReportPolicy.shouldReportFailedDownload(download.getState(),
+                download.getAnnouncedContentLength(), file.getActualChecksum().getContentLength(),
```
cpesch, line 52: that name shouldReportFailedDownload() is weird - is tests wehre it should report a sucessful download

cpesch:
🤖 Patcher pushed a fix to `agent/builder/382-download-checksumsender-reports-the-chec`.

cpesch:

### Panel

- **correctness** — request-changes · confidence 0.90
  The predicate's fail-closed design is sound, but the new test class cannot pass as written: differingLastModifiedIsNotReported picks two timestamps inside the same second, which the policy's own second-precision rounding treats as equal, so it directly contradicts lastModifiedComparisonIgnoresMillisecondPrecision a few lines below — and nothing in this PR was ever compiled or run to catch it. Two lesser risks: a removed Result constructor overload may have a caller outside the shown hunks, and the fix's positive half (genuine rebuild actually reports) depends on Checksum lastModified wiring that is neither in the diff nor asserted by the new wiring test.

### download/src/main/java/slash/navigation/download/performer/GetPerformer.java

- **minor** [correctness] line 268 — The 3-arg Result(Get, boolean, Long) constructor was removed in favour of this 4-arg form, and the only migrated caller visible in the diff is download()'s full-body path. If resume() — whose body is outside the shown hunks — constructs a Result with a Long third argument, the download module no longer compiles, and per the PR body compilation never ran (no JDK on PATH, factory-verify rejected the repo), so nothing confirms the overload had no other callers.  _(correctness)_

```suggestion
Compile the download module, or confirm via grep that resume() constructs Result only through the surviving (Get, boolean) and (Get, boolean, boolean) overloads; if resume() did carry a lastModified, decide explicitly whether resumed downloads should keep it (their null contentLength already fails the predicate closed).
```

### download/src/test/java/slash/navigation/download/ChecksumReportPolicyTest.java

- **major** [correctness] line 86 — This test is guaranteed red: LAST_MODIFIED (1_700_000_000_123) and LAST_MODIFIED_OTHER (1_700_000_000_456) fall within the same second, so after roundMillisecondsToSecondPrecision both sides are equal and isReportableChecksum returns true (state is ChecksumError and the lengths match), making assertFalse fail — under floor, round-half-up or ceil alike, any second-precision rounding collapses the two constants to the same value. The test therefore contradicts its sibling lastModifiedComparisonIgnoresMillisecondPrecision (line 123), which asserts the same-second tolerance yields true; the two cannot both pass, and per the PR body no compile or test run ever happened to surface it.  _(correctness)_

```suggestion
Move LAST_MODIFIED_OTHER into a different second (e.g. 1_700_000_001_456L at line 45) so the test exercises a genuinely different build timestamp rather than sub-second noise, which the policy is designed to tolerate.
```

### route-converter-gui/src/main/java/slash/navigation/converter/gui/helpers/ChecksumSender.java

- **minor** [correctness] line 54 — The positive half of the fix now hinges on file.getActualChecksum().getLastModified() mirroring the response's announced Last-Modified for a ChecksumError download, but the code that populates the actual checksum is outside the diff and the new wiring test asserts only the announced content length (its expected Checksum is even built with a null lastModified). If the actual checksum carries a null or differently derived lastModified, the predicate fails closed for every genuine rebuild and the catalog never relearns — the bug the issue wants fixed on that side would silently stop working, with no red test.  _(correctness)_

```suggestion
Extend the wiring test to serve a Last-Modified header and assert end-to-end that a completed-but-mismatching download produces matching announced/actual lastModified values, so the genuine-rebuild path is exercised through the real performer rather than only through the pure predicate unit test.
```

---

cpesch:
🤖 Patcher pushed a fix to `agent/builder/382-download-checksumsender-reports-the-chec`.

cpesch:

### Panel

  The predicate and its wiring are largely sound and fail closed, but the PR ships unverified (no JDK/Maven in the builder session), and the rewritten `Result` record removed a convenience constructor whose remaining callers are outside this diff. Two smaller gaps: the resume path can never satisfy the new policy, and the Last-Modified leg compares a value against its own provenance.

### download/src/main/java/slash/navigation/download/ChecksumReportPolicy.java

- **minor** [correctness] line 66 — The Last-Modified leg compares the announced header against the actual checksum's mtime, which `copy()` stamps from that same header (the PR's own end-to-end test documents this), so both sides share provenance and the check cannot independently reject a download; `differingLastModifiedIsNotReported` exercises a state unreachable at runtime. Only the Content-Length leg does real work, yet the javadoc claims the mtime proves the file is the announced build.  _(correctness)_

```suggestion
Keep the leg only as an explicit tripwire for a failed `File.setLastModified` (rename/comment it that way), or drop it — either way the javadoc should not describe an assurance the comparison cannot deliver.
```

### download/src/main/java/slash/navigation/download/performer/GetPerformer.java

- **major** [correctness] line 268 — The convenience constructor `Result(Get, boolean, Long)` was removed while rewriting the record; `download()`'s call site was migrated, but any other caller in this unchanged file — most plausibly `resume()`, whose body is not in this diff — that still passes `(request, success, lastModified)` no longer resolves. The PR is explicitly UNVERIFIED (no JDK, broker rejects the repo), so a compile break would not have been caught before merge.  _(correctness)_

```suggestion
Compile `download` before merge, or defensively re-add `public Result(Get request, boolean success, Long lastModified) { this(request, success, false, lastModified, null); }`. Note the resume path must NOT forward a 206's Content-Length into that slot — a range length would misrepresent the whole resource and could make the policy accept a partial file.
```
- **minor** [correctness] line 151 — Only `download()`'s full-body path populates `announcedContentLength`; a transfer completed via `resume()` reaches `result.success` with a null contentLength (a 206's Content-Length describes only the remaining range), so `isReportableChecksum` always returns false on that path. A genuine upstream rebuild delivered as a resumed transfer is silently never reported — the #382 fix is intentionally conservative, but this case regresses relative to the stated intent that complete-and-validated failures still report.  _(correctness)_

```suggestion
Derive the whole-resource length on the resume path (pre-existing temp-file size plus the served range, or the expected checksum's content length), or state in the ChecksumReportPolicy javadoc that resumed downloads never report a failed-download checksum.
```

---

cpesch:

**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-27259` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-27259).

PR #390, opened 2026-09-10
test(mapsforge-mapview): TileRendererLayer lifecycle test for remove/re-add and theme switch

### LINKED ISSUE #384: [mapsforge-mapview] lifecycle test: TileRendererLayer keeps rendering after remove + re-add and theme switch
## Test: TileRendererLayer lifecycle in the map view — a layer must keep rendering after remove + re-add and after a theme switch

**Source:** maintainer session 2026-09-06, after #381. Goal: catch the whole class of "layer is attached and alive but never renders another tile" regressions with a headless test, including future mapsforge upgrades (mapsforge/mapsforge#1817 is unfixed upstream). This spec is fully resolved; implement without further questions.

### What is established

- #376/#381: the world-map background was silently dead for weeks because `handleBackground()` removed and re-added the layer, and mapsforge's `TileRendererLayer` keeps its `MapWorkerPool` on the `JobQueue` of the *first* `setDisplayModel()`. No test caught it: `BackgroundMapAttachmentTest` covers pure predicates, `MapsforgeMapViewTest` cannot build a `LayerManager`, `ReattachableTileRendererLayerTest` asserts queue identity via reflection, not rendering.
- A 20-line harness proved the bug in seconds: construct `Layers`, add the layer, call `layer.draw(...)` once, wait, count rendered tiles. That is the behavioural check this issue turns into a test.
- **No `.map` fixture is needed.** Probed 2026-09-06: a `TileRendererLayer` over an empty `new MultiMapDataStore()` with `MapsforgeThemes.OSMARENDER` at zoom 7, 1024×768 canvas, renders and caches **15 tiles** within 3 s (empty transparent bitmaps go through `DatabaseRenderer.executeJob` → `tileCache.put` like real ones). The lifecycle bug is independent of map content.
- `org.mapsforge.map.layer.Layers` has a package-private constructor, so the test class lives in package `org.mapsforge.map.layer` under `mapsforge-mapview/src/test/java` (split package on the classpath is fine; the module is not JPMS).
- Rendering happens on `MapWorkerPool` threads; `Layers.add()` starts them, `Layers.remove()` stops them (`onRemove` → `MapWorkerPool.stop()`).

### Scope

One new test class in `mapsforge-mapview`. No production code changes. Out of scope: testing `MapsforgeMapView.handleBackground()` itself (needs a `MapView`), the `DefaultTileLayerFactory` `.map`-file path, MBTiles/download layers.

### Files

- `mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java` (new)

1. **Setup per test:** `DisplayModel dm = new DisplayModel()`, `MapViewPosition mvp` centred at `LatLong(54.0, 10.0)` zoom 7, `InMemoryTileCache(64)` (first level only, no filesystem), `GraphicFactory gf = AwtGraphicFactory.INSTANCE`, `Layers layers = new Layers(() -> {}, dm)`, canvas `gf.createCanvas()` over `gf.createBitmap(1024, 768)`, bounding box and top-left via `MapPositionUtil.getBoundingBox(mp, Rotation.NULL_ROTATION, dm.getTileSize(), new Dimension(1024, 768), 0.5f, 0.5f)` and `MapPositionUtil.getTopLeftPoint(mp, dim, tileSize)`.
2. **Layer under test:** `slash.navigation.mapview.mapsforge.tiles.ReattachableTileRendererLayer(cache, new MultiMapDataStore(), mvp, true, true, true, gf, null)` with `setXmlRenderTheme(MapsforgeThemes.OSMARENDER)`, i.e. exactly what `DefaultTileLayerFactory.createTileRendererLayer()` builds minus the `MapFile`.
3. **"Rendered" assertion:** after `layer.draw(bbox, zoom, canvas, topLeft, Rotation.NULL_ROTATION)`, poll every 100 ms for up to 5 s until `cache.containsKey(job)` is true for **every** job of `LayerUtil.getTilePositions(bbox, zoom, topLeft, tileSize)` (build jobs the way `TileLayer.draw()` does: `new RendererJob(tile, layer.getMapDataStore(), layer.getRenderThemeFuture(), dm, 1f, true, false)`). Fail with the count of missing tiles in the message. No filesystem, no sleeps beyond the poll.
4. **Cases:**
   - `rendersAfterSingleAdd` — add, draw, all tiles cached.
   - `rendersAfterRemoveAndReAdd` — add, `layers.remove(layer)`, `layers.add(0, layer)`, draw, all tiles cached. Comment: "with a plain `TileRendererLayer` this fails (mapsforge/mapsforge#1817): zero tiles ever render".
   - `rendersAfterThemeSwitchWhileAttached` — add, draw + wait, `cache.purge()`, `layer.setXmlRenderTheme(MapsforgeThemes.DEFAULT)`, draw again, all tiles cached (a new `RenderThemeFuture` must be picked up by new jobs).
   - `rendersAfterRemoveReAddWithJobsInFlight` — add, draw (do not wait), immediately remove + re-add, draw again, all tiles cached within the poll window. This is the startup race from #376 (tiles interrupted by the rebuild are re-queued and finished).
5. **Teardown (`@After`):** `layers.remove(layer)` if still attached, `layer.onDestroy()` is protected → call `layer.setDisplayModel(null)` to stop the pool, then `cache.destroy()`. Tests must leave no `MapWorkerPool` threads behind (assert `Thread.getAllStackTraces()` has no thread whose stack contains `MapWorkerPool.run` after teardown; poll up to 2 s since executors shut down asynchronously).
6. Runtime budget: whole class under 10 s; it runs in the normal surefire unit phase (name ends in `Test`).

### Testable seam

The test *is* the seam; no production helper. Assertion helper inside the test class:
```java
private static void assertAllTilesRendered(TileCache cache, TileLayer<RendererJob> layer, List<TilePosition> positions, DisplayModel dm) throws InterruptedException
```

### Acceptance criteria

- `mvn -pl mapsforge-mapview test` green, class runtime < 10 s, no leaked `MapWorkerPool` threads.
- Mutation check by the builder (document the result in the PR): replace `ReattachableTileRendererLayer` with `TileRendererLayer` in `rendersAfterRemoveAndReAdd` → the test fails with 0 of 15 tiles rendered; revert.
- Second mutation check: in `ReattachableTileRendererLayer.setDisplayModel()` drop the `jobQueue = firstJobQueue` restore → same failure; revert.

### Related

#376, #381, mapsforge/mapsforge#1817, `ReattachableTileRendererLayerTest`.

cpesch:
## Summary
- New headless test `TileRendererLayerLifecycleTest` (package `org.mapsforge.map.layer`, to reach `Layers`' package-private constructor) proves a `TileRendererLayer` keeps rendering tiles after `Layers.remove()`+`add()` and after a theme switch while attached.
- No production code changes; scope is limited to `mapsforge-mapview`, per the issue.

## Test plan
- [x] `mvn -pl mapsforge-mapview test` green: `Tests run: 4, Failures: 0, Errors: 0, Time elapsed: 5.541 s` (< 10s budget)
- [x] Mutation check 1: swapped `ReattachableTileRendererLayer` for a plain `TileRendererLayer` in `rendersAfterRemoveAndReAdd` → fails with `20 of 20 tiles not rendered` (mapsforge/mapsforge#1817 reproduced); reverted.
- [x] Mutation check 2: dropped the `jobQueue = firstJobQueue` restore in `ReattachableTileRendererLayer.setDisplayModel()` → same failure, `20 of 20 tiles not rendered`; reverted.
- [x] No leaked `MapWorkerPool` threads after teardown (asserted in `@After`, polled up to 2s).

Closes #384

### DIFF
--- a/mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java
+++ b/mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java
@@ -1,0 +1,204 @@
+/*
+    This file is part of RouteConverter.
+
+    RouteConverter is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    RouteConverter is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with RouteConverter; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+    Copyright (C) 2007 Christian Pesch. All Rights Reserved.
+*/
+package org.mapsforge.map.layer;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.mapsforge.core.graphics.Bitmap;
+import org.mapsforge.core.graphics.Canvas;
+import org.mapsforge.core.graphics.GraphicFactory;
+import org.mapsforge.core.model.BoundingBox;
+import org.mapsforge.core.model.Dimension;
+import org.mapsforge.core.model.LatLong;
+import org.mapsforge.core.model.MapPosition;
+import org.mapsforge.core.model.Point;
+import org.mapsforge.core.model.Rotation;
+import org.mapsforge.map.awt.graphics.AwtGraphicFactory;
+import org.mapsforge.map.datastore.MultiMapDataStore;
+import org.mapsforge.map.layer.cache.InMemoryTileCache;
+import org.mapsforge.map.layer.cache.TileCache;
+import org.mapsforge.map.layer.renderer.MapWorkerPool;
+import org.mapsforge.map.layer.renderer.RendererJob;
+import org.mapsforge.map.layer.renderer.TileRendererLayer;
+import org.mapsforge.map.model.DisplayModel;
+import org.mapsforge.map.model.MapViewPosition;
+import org.mapsforge.map.rendertheme.internal.MapsforgeThemes;
+import org.mapsforge.map.util.LayerUtil;
+import org.mapsforge.map.util.MapPositionUtil;
+import slash.navigation.mapview.mapsforge.tiles.ReattachableTileRendererLayer;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Proves a {@link TileRendererLayer} keeps rendering after it has been removed from and re-added to
+ * {@link Layers}, and after a theme switch while attached.
+ * <p>
+ * A plain {@link TileRendererLayer} fails {@link #rendersAfterRemoveAndReAdd()} (mapsforge/mapsforge#1817):
+ * its {@link MapWorkerPool} keeps waiting on the {@link org.mapsforge.map.layer.queue.JobQueue} of the
+ * first {@code setDisplayModel()} call, while {@link TileLayer#draw} queues jobs into the fresh queue
+ * {@code Layers.add()} installs on every add. {@link ReattachableTileRendererLayer} works around this.
+ */
+public class TileRendererLayerLifecycleTest {
+    private static final int WIDTH = 1024;
+    private static final int HEIGHT = 768;
+    private static final GraphicFactory GRAPHIC_FACTORY = AwtGraphicFactory.INSTANCE;
+
+    private DisplayModel displayModel;
+    private InMemoryTileCache cache;
+    private Layers layers;
+    private ReattachableTileRendererLayer layer;
+    private Canvas canvas;
+    private BoundingBox boundingBox;
+    private byte zoomLevel;
+    private Point topLeftPoint;
+    private List<TilePosition> tilePositions;
+
+    @Before
+    public void setUp() {
+        displayModel = new DisplayModel();
+        MapViewPosition mapViewPosition = new MapViewPosition(displayModel);
+        mapViewPosition.setMapPosition(new MapPosition(new LatLong(54.0, 10.0), (byte) 7), false);
+
+        cache = new InMemoryTileCache(64);
+        layers = new Layers(() -> {
+        }, displayModel);
+
+        layer = new ReattachableTileRendererLayer(cache, new MultiMapDataStore(), mapViewPosition,
+                true, true, true, GRAPHIC_FACTORY, null);
+        layer.setXmlRenderTheme(MapsforgeThemes.OSMARENDER);
+
+        Dimension dimension = new Dimension(WIDTH, HEIGHT);
+        MapPosition mapPosition = mapViewPosition.getMapPosition();
+        boundingBox = MapPositionUtil.getBoundingBox(mapPosition, Rotation.NULL_ROTATION,
+                displayModel.getTileSize(), dimension, 0.5f, 0.5f);
+        zoomLevel = mapPosition.zoomLevel;
+        topLeftPoint = MapPositionUtil.getTopLeftPoint(mapPosition, dimension, displayModel.getTileSize());
+        tilePositions = LayerUtil.getTilePositions(boundingBox, zoomLevel, topLeftPoint, displayModel.getTileSize());
+
+        Bitmap bitmap = GRAPHIC_FACTORY.createBitmap(WIDTH, HEIGHT);
+        canvas = GRAPHIC_FACTORY.createCanvas();
+        canvas.setBitmap(bitmap);
+    }
+
+    @After
+    public void tearDown() throws InterruptedException {
+        if (layers.contains(layer))
+            layers.remove(layer);
+        layer.setDisplayModel(null);
+        cache.destroy();
+        assertNoLeakedMapWorkerPoolThreads();
+    }
+
+    @Test
+    public void rendersAfterSingleAdd() throws InterruptedException {
+        layers.add(layer);
+
+        draw();
+
+        assertAllTilesRendered();
+    }
+
+    @Test
+    public void rendersAfterRemoveAndReAdd() throws InterruptedException {
+        layers.add(layer);
+        layers.remove(layer);
+        layers.add(0, layer);
+
+        draw();
+
+        // with a plain TileRendererLayer this fails (mapsforge/mapsforge#1817): zero tiles ever render
+        assertAllTilesRendered();
+    }
+
+    @Test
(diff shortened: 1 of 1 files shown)


mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java
```diff
@@ -0,0 +1,204 @@
...
+    @Test
+    public void rendersAfterRemoveReAddWithJobsInFlight() throws InterruptedException {
+        layers.add(layer);
+        draw(); // do not wait for the jobs to finish before the rebuild
+
+        layers.remove(layer);
+        layers.add(0, layer);
+        draw();
+
+        assertAllTilesRendered();
```
cpesch, line 158: **minor** [correctness] rendersAfterRemoveReAddWithJobsInFlight cannot detect the mapsforge#1817 regression it is named for: assertAllTilesRendered probes cache.containsKey for RendererJob keys identical to the jobs enqueued by the pre-rebuild draw at line 152 (same tiles, same map data store, same render theme future, same displayModel), the cache is never purged in this test, and — per the class's own javadoc — the plain layer's orphaned MapWorkerPool keeps draining the first JobQueue, so pre-rebuild work alone satisfies the assertion even when the re-added layer never renders again (contrast rendersAfterRemoveAndReAdd, where no draw precedes the rebuild, which the PR's mutation checks validated).
```suggestion
Either drop this test in favour of rendersAfterRemoveAndReAdd, or make the assertion provable only by post-rebuild rendering: purge the cache after layers.add(0, layer) and before the second draw (the orphaned worker has already drained the pre-rebuild queue, so nothing re-renders on a broken layer and the assert fails), or change the map position before the second draw so the probed tiles were never enqueued pre-rebuild.
```

cpesch (commented):

### Panel

- **correctness** — comment-only · confidence 0.84
  Test-only diff, complete slice visible; the no-production-code claim matches files_changed and the two cited mutation checks anchor the key regression test to a real failure mode. One minor finding: rendersAfterRemoveReAddWithJobsInFlight cannot fail for the regression its name claims, because nothing clears the cache between its pre-rebuild draw and the assertion it probes.

### mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java

- **minor** [correctness] line 158 — rendersAfterRemoveReAddWithJobsInFlight cannot detect the mapsforge#1817 regression it is named for: assertAllTilesRendered probes cache.containsKey for RendererJob keys identical to the jobs enqueued by the pre-rebuild draw at line 152 (same tiles, same map data store, same render theme future, same displayModel), the cache is never purged in this test, and — per the class's own javadoc — the plain layer's orphaned MapWorkerPool keeps draining the first JobQueue, so pre-rebuild work alone satisfies the assertion even when the re-added layer never renders again (contrast rendersAfterRemoveAndReAdd, where no draw precedes the rebuild, which the PR's mutation checks validated).  _(correctness)_

```suggestion
Either drop this test in favour of rendersAfterRemoveAndReAdd, or make the assertion provable only by post-rebuild rendering: purge the cache after layers.add(0, layer) and before the second draw (the orphaned worker has already drained the pre-rebuild queue, so nothing re-renders on a broken layer and the assert fails), or change the map position before the second draw so the probed tiles were never enqueued pre-rebuild.
```

---

cpesch:

**Confidence:** 0

Full dry-run detail on the Gitea branch `agent/patcher/dryrun-28386` (https://git.routeconverter.com/rc/rc-meta/src/branch/agent/patcher/dryrun-28386).

mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java
```diff
@@ -0,0 +1,206 @@
...
+            if (!anyMapWorkerPoolThreadRunning())
+                return;
+            Thread.sleep(100);
+        } while (System.currentTimeMillis() < deadline);
+
+        assertTrue("MapWorkerPool thread still running after teardown", !anyMapWorkerPoolThreadRunning());
+    }
+
+    private static boolean anyMapWorkerPoolThreadRunning() {
+        for (Map.Entry<Thread, StackTraceElement[]> entry : Thread.getAllStackTraces().entrySet()) {
```
cpesch, line 198: **nit** [correctness] The leak guard scans all threads JVM-wide, so a MapWorkerPool worker leaked by any other test class in the same (by default reused) surefire fork fails this test's @After even though this test cleaned up after itself; attribution lands on this class regardless of who leaked.
```suggestion
Snapshot the live MapWorkerPool thread set in setUp and fail only on threads that appeared during this test, or match on a thread-name tag specific to the pool this layer starts.
```

mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java
```diff
@@ -0,0 +1,206 @@
...
+            if (missing.isEmpty())
+                return;
+            Thread.sleep(100);
+        } while (System.currentTimeMillis() < deadline);
+
+        fail(missing.size() + " of " + tilePositions.size() + " tiles not rendered");
+    }
+
+    private static void assertNoLeakedMapWorkerPoolThreads() throws InterruptedException {
+        long deadline = System.currentTimeMillis() + 2000;
```
cpesch, line 187: **nit** [correctness] The 2s wind-down budget is asymmetric with the 15s render budget: if MapWorkerPool.stop() only sets a stop flag (not visible in this diff), a worker caught mid-tile-render can legitimately stay inside run() past 2s on a loaded CI runner and red the teardown although the code under test is fine.
```suggestion
Give the wind-down poll the same headroom as assertAllTilesRendered, or await pool termination/interrupt explicitly if the mapsforge API exposes it, before asserting.
```

cpesch (commented):

### Panel

- **correctness** — comment-only · confidence 0.88
  Test-only change: the lifecycle scenarios, the lookup-job parity with the worker's job construction, and the deadline-poll waits all check out, and no production surface is touched. Two nits, both on the teardown thread-leak assertion: its JVM-wide scan can blame a sibling class's leak on this test, and its 2s wind-down window is far tighter than the 15s render budget it shares a runner with.

### mapsforge-mapview/src/test/java/org/mapsforge/map/layer/TileRendererLayerLifecycleTest.java

- **nit** [correctness] line 187 — The 2s wind-down budget is asymmetric with the 15s render budget: if MapWorkerPool.stop() only sets a stop flag (not visible in this diff), a worker caught mid-tile-render can legitimately stay inside run() past 2s on a loaded CI runner and red the teardown although the code under test is fine.  _(correctness)_

```suggestion
Give the wind-down poll the same headroom as assertAllTilesRendered, or await pool termination/interrupt explicitly if the mapsforge API exposes it, before asserting.
```
- **nit** [correctness] line 198 — The leak guard scans all threads JVM-wide, so a MapWorkerPool worker leaked by any other test class in the same (by default reused) surefire fork fails this test's @After even though this test cleaned up after itself; attribution lands on this class regardless of who leaked.  _(correctness)_

```suggestion
Snapshot the live MapWorkerPool thread set in setUp and fail only on threads that appeared during this test, or match on a thread-name tag specific to the pool this layer starts.
```

---

PR #391, opened 2026-09-10
fix(navigation-formats): fix GoogleTimelineFormat exception and dropped points

cpesch:
## Summary
- `Wgs84Route` had no `asGoogleTimelineFormat()`, so `NavigationFormatConverter`'s reflection-based round-trip threw `IOException: Cannot call asGoogleTimelineFormat()` on every load. Added it following the existing `as...Format()` convention.
- The on-device Google Timeline export's `timelinePath` points use a bare `"point": "geo:lat,lng"` key instead of `"location"`, so all track points were silently dropped, leaving only Visit waypoints. Added the fallback.

Repro'd against a real ~2900-entry `location-history.json` export: before the fix it threw; after, it loads 505 routes (day tracks + visits) with no data loss.

## Test plan
- [x] Reproduced original exception and fixed it against a real export
- [x] `GoogleTimelineFormatTest` passes
- [x] Manually loaded the file in the running RouteConverter GUI

### DIFF
--- a/navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
+++ b/navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
@@ -335,6 +335,10 @@ public class GoogleTimelineFormat extends SimpleFormat<Wgs84Route> {
                                         Activity precedingActivity, boolean isIOS) {
         JsonNode locationNode = point.path("location");
         if (locationNode == null || locationNode.isMissingNode()) {
+            // on-device export shape: bare "point": "geo:lat,lng" instead of "location"
+            locationNode = point.path("point");
+        }
+        if (locationNode == null || locationNode.isMissingNode()) {
             return null;
         }
 
--- a/navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
+++ b/navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
@@ -53,6 +53,12 @@ public class GoogleTimelineFormat extends SimpleFormat<Wgs84Route> {
     private static final String GEO_PREFIX = "geo:";
     private static final Pattern COORDINATE_PATTERN = Pattern.compile("(-?\\d+\\.\\d+)°?,?\\s*(-?\\d+\\.\\d+)°?");
 
+    // Calendar-day grouping is an artifact of the export, not of the trip: an overnight
+    // stay splits one continuous track at midnight even though the last point of one day
+    // and the first point of the next are (almost) the same place. Bridge that gap instead
+    // of emitting two routes that visibly touch.
+    private static final double ADJACENT_DAY_MERGE_THRESHOLD_METERS = 50.0;
+
     public String getExtension() {
         return ".json";
     }
@@ -169,7 +175,13 @@ public class GoogleTimelineFormat extends SimpleFormat<Wgs84Route> {
             }
         }
 
-        // Emit day tracks in ascending date order
+        // Emit day tracks in ascending date order, merging a day into the still-open
+        // previous track when its first point is (almost) the same place as the previous
+        // track's last point - see ADJACENT_DAY_MERGE_THRESHOLD_METERS above.
+        List<Wgs84Position> pendingPoints = null;
+        CompactCalendar pendingFirstDay = null;
+        CompactCalendar pendingLastDay = null;
+
         for (Map.Entry<CompactCalendar, List<Wgs84Position>> entry : pointsByDay.entrySet()) {
             List<Wgs84Position> dayPoints = entry.getValue();
             // Sort by timestamp, stable for ties
@@ -182,9 +194,19 @@ public class GoogleTimelineFormat extends SimpleFormat<Wgs84Route> {
                 return 0;
             });
 
-            String routeName = formatDateName(entry.getKey());
-            context.appendRoute(new Wgs84Route(this, Track, routeName, dayPoints));
+            if (pendingPoints != null && isSameLocation(pendingPoints.get(pendingPoints.size() - 1), dayPoints.get(0))) {
+                pendingPoints.addAll(dayPoints);
+                pendingLastDay = entry.getKey();
+            } else {
+                if (pendingPoints != null)
+                    context.appendRoute(new Wgs84Route(this, Track, formatRouteName(pendingFirstDay, pendingLastDay), pendingPoints));
+                pendingPoints = dayPoints;
+                pendingFirstDay = entry.getKey();
+                pendingLastDay = entry.getKey();
+            }
         }
+        if (pendingPoints != null)
+            context.appendRoute(new Wgs84Route(this, Track, formatRouteName(pendingFirstDay, pendingLastDay), pendingPoints));
 
         // Process visits into a single Waypoints route
         List<Wgs84Position> visitPositions = new ArrayList<>();
@@ -277,6 +299,17 @@ public class GoogleTimelineFormat extends SimpleFormat<Wgs84Route> {
         int month = calendar.get(MONTH) + 1; // Calendar.MONTH is 0-based
         int dayOfMonth = calendar.get(DAY_OF_MONTH);
         return String.format("%04d-%02d-%02d", year, month, dayOfMonth);
+    }
+
+    private String formatRouteName(CompactCalendar firstDay, CompactCalendar lastDay) {
+        String firstName = formatDateName(firstDay);
+        String lastName = formatDateName(lastDay);
+        return firstName.equals(lastName) ? firstName : firstName + " – " + lastName;
+    }
+
+    private boolean isSameLocation(Wgs84Position a, Wgs84Position b) {
+        Double distance = b.calculateDistance(a);
+        return distance != null && distance <= ADJACENT_DAY_MERGE_THRESHOLD_METERS;
     }
 
     private Activity parseActivity(JsonNode segment, boolean isAndroid) {
--- a/navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
+++ b/navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
@@ -30,6 +30,7 @@ import slash.navigation.common.NavigationPosition;
 import java.io.*;
 import java.util.*;
 import java.util.logging.Logger;
+import java.util.prefs.Preferences;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
```diff
@@ -334,6 +334,10 @@ private List<Wgs84Position> parseTimelinePath(JsonNode segment, Activity precedi
     private Wgs84Position parsePathPoint(JsonNode point, CompactCalendar segmentStart,
                                         Activity precedingActivity, boolean isIOS) {
         JsonNode locationNode = point.path("location");
+        if (locationNode == null || locationNode.isMissingNode()) {
+            // on-device export shape: bare "point": "geo:lat,lng" instead of "location"
+            locationNode = point.path("point");
```
cpesch, line 377: **minor** [correctness] The fallback assigns point.path("point") to locationNode, but the new comment itself describes that value as a "geo:lat,lng" string while the code downstream of this method was written for the "location" object shape — and the hunk ends before that consumer, so I cannot confirm the string form is actually parsed. If it is not, track points no longer get dropped but resurface at (0,0) (missing children on a TextNode resolve to 0.0), silently replacing data loss with data corruption. Secondary edge the diff newly reaches: a JSON-null "point" yields a NullNode, which passes both isMissingNode() guards and flows to the same consumer.
```suggestion
Verify the parsing branch after line 343 handles a TextNode geo: value (the PR's real-export GUI check likely already covers this); a test asserting parsed path-point coordinates are non-zero would pin it down. Optionally gate the fallback on locationNode.isObject() || locationNode.isTextual() and parse the geo: URI explicitly, returning null otherwise.
```

cpesch (commented):

### Panel

- **correctness** — comment-only · confidence 0.90
  Both changed files are fully covered. The asGoogleTimelineFormat() addition mirrors the neighbouring asGeoJsonFormat()/asMicrosoftExcel97Format() pattern with no runtime risk I can see, and the parsePathPoint() fallback preserves the old drop behaviour when neither key exists; one open question remains about the shape of the fallback value versus what the downstream location parser expects.

### navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java

- **minor** [correctness] line 339 — The fallback assigns point.path("point") to locationNode, but the new comment itself describes that value as a "geo:lat,lng" string while the code downstream of this method was written for the "location" object shape — and the hunk ends before that consumer, so I cannot confirm the string form is actually parsed. If it is not, track points no longer get dropped but resurface at (0,0) (missing children on a TextNode resolve to 0.0), silently replacing data loss with data corruption. Secondary edge the diff newly reaches: a JSON-null "point" yields a NullNode, which passes both isMissingNode() guards and flows to the same consumer.  _(correctness)_

```suggestion
Verify the parsing branch after line 343 handles a TextNode geo: value (the PR's real-export GUI check likely already covers this); a test asserting parsed path-point coordinates are non-zero would pin it down. Optionally gate the fallback on locationNode.isObject() || locationNode.isTextual() and parse the geo: URI explicitly, returning null otherwise.
```

---

navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
```diff
@@ -182,9 +194,19 @@ public void read(BufferedReader reader, String encoding, ParserContext<Wgs84Rout
...
 
-            String routeName = formatDateName(entry.getKey());
-            context.appendRoute(new Wgs84Route(this, Track, routeName, dayPoints));
+            if (pendingPoints != null && isSameLocation(pendingPoints.get(pendingPoints.size() - 1), dayPoints.get(0))) {
+                pendingPoints.addAll(dayPoints);
+                pendingLastDay = entry.getKey();
+            } else {
+                if (pendingPoints != null)
+                    context.appendRoute(new Wgs84Route(this, Track, formatRouteName(pendingFirstDay, pendingLastDay), pendingPoints));
+                pendingPoints = dayPoints;
```
cpesch, line 207: **minor** [correctness] The merge path indexes dayPoints.get(0) and pendingPoints.get(pendingPoints.size() - 1) (line 197) without an emptiness check, so an empty day bucket would throw IndexOutOfBoundsException where the previous code emitted an empty route instead. Whether an empty bucket is constructible depends on how pointsByDay is populated, which is not visible in this diff; if buckets are only created when a position is added this is unreachable, but the new code silently assumes it.
```suggestion
Skip empty buckets (continue if dayPoints.isEmpty()), or state/verify in the guard comment that pointsByDay is only populated via computeIfAbsent-on-add so empty buckets cannot exist.
```

cpesch (commented):

### Panel

  The fix logic is order-safe (ascending day iteration, per-day timestamp sort, fail-safe on null distance) and the new asGoogleTimelineFormat follows the sibling asXxxFormat pattern. Two minor gaps from this lens: the adjacent-day merge has no calendar-adjacency guard, and the merge path assumes non-empty day buckets. The textual geo: node handling downstream of the new parsePathPoint fallback is not visible in this diff, but the PR's stated end-to-end repro against a real on-device export corroborates it, so no finding is raised there.

### navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java

- **minor** [correctness] line 197 — The merge condition checks only positional proximity (50 m) between the pending run's last point and the next day's first point, with no requirement that the two days be calendar-adjacent. A trip that ends at a place and a later trip that starts there (gap days in between producing only visits, or a revisited hotel weeks apart) collapse into one route whose name spans the whole range and whose timestamps jump by the gap. At runtime this silently merges distinct trips into a single track.  _(correctness)_

```suggestion
Require adjacency in addition to proximity: only merge when entry.getKey() is exactly one day after pendingLastDay. That still bridges the midnight split the new comment targets and stops indefinite chaining across date gaps.
```
- **minor** [correctness] line 203 — The merge path indexes dayPoints.get(0) and pendingPoints.get(pendingPoints.size() - 1) (line 197) without an emptiness check, so an empty day bucket would throw IndexOutOfBoundsException where the previous code emitted an empty route instead. Whether an empty bucket is constructible depends on how pointsByDay is populated, which is not visible in this diff; if buckets are only created when a position is added this is unreachable, but the new code silently assumes it.  _(correctness)_

```suggestion
Skip empty buckets (continue if dayPoints.isEmpty()), or state/verify in the guard comment that pointsByDay is only populated via computeIfAbsent-on-add so empty buckets cannot exist.
```

---

navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
```diff
@@ -182,9 +194,19 @@ public void read(BufferedReader reader, String encoding, ParserContext<Wgs84Rout
                 return 0;
             });
 
-            String routeName = formatDateName(entry.getKey());
-            context.appendRoute(new Wgs84Route(this, Track, routeName, dayPoints));
+            if (pendingPoints != null && isSameLocation(pendingPoints.get(pendingPoints.size() - 1), dayPoints.get(0))) {
```
cpesch, line 201: **minor** [correctness] The merge condition checks only positional proximity (50 m) between the pending run's last point and the next day's first point, with no requirement that the two days be calendar-adjacent. A trip that ends at a place and a later trip that starts there (gap days in between producing only visits, or a revisited hotel weeks apart) collapse into one route whose name spans the whole range and whose timestamps jump by the gap. At runtime this silently merges distinct trips into a single track.
```suggestion
Require adjacency in addition to proximity: only merge when entry.getKey() is exactly one day after pendingLastDay. That still bridges the midnight split the new comment targets and stops indefinite chaining across date gaps.
```

navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
```diff
@@ -182,9 +198,19 @@ public void read(BufferedReader reader, String encoding, ParserContext<Wgs84Rout
...
```
cpesch, line 201: **minor** [correctness] The merge condition tests distance only, not calendar adjacency: a day whose first point sits within the threshold of the pending track's last point is fused into it even when several calendar days elapsed, so a stay recorded on non-consecutive days (same hotel, intervening days with no recorded track) silently becomes one Track with a temporal discontinuity.
```suggestion
Additionally require the two days to be exactly one calendar day apart (compare pendingLastDay with entry.getKey()) before merging, so only true overnight splits are bridged.
```

navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java
```diff
@@ -279,6 +305,18 @@ private String formatDateName(CompactCalendar day) {
         return String.format("%04d-%02d-%02d", year, month, dayOfMonth);
     }
 
+    private String formatRouteName(CompactCalendar firstDay, CompactCalendar lastDay) {
+        String firstName = formatDateName(firstDay);
+        String lastName = formatDateName(lastDay);
+        return firstName.equals(lastName) ? firstName : firstName + " – " + lastName;
+    }
+
+    private boolean isSameLocation(Wgs84Position a, Wgs84Position b) {
```
cpesch, line 314: **nit** [correctness] Merged route names now carry a non-ASCII en dash (U+2013); previously Timeline route names were pure ASCII dates, and some targets reachable via NavigationFormatConverter serialize with restricted charsets where the dash could be mangled. The writers are not visible in this diff, so I cannot name a concrete failing one.

cpesch (commented):

### Panel

- **correctness** — comment-only · confidence 0.86
  The asGoogleTimelineFormat() reflection hook and the "point"-key fallback follow the existing sibling conventions and I found no blocker in the changed lines; the merge loop emits each group exactly once and the single-day name case reproduces the old output. One real edge case the diff newly reaches: the day-merge guard compares distance only, never calendar adjacency, so non-consecutive days at the same overnight location fuse into a single track spanning the gap. Diff coverage is complete; nothing was omitted or truncated.

### navigation-formats/src/main/java/slash/navigation/googletimeline/GoogleTimelineFormat.java

- **minor** [correctness] line 201 — The merge condition tests distance only, not calendar adjacency: a day whose first point sits within the threshold of the pending track's last point is fused into it even when several calendar days elapsed, so a stay recorded on non-consecutive days (same hotel, intervening days with no recorded track) silently becomes one Track with a temporal discontinuity.  _(correctness)_

```suggestion
Additionally require the two days to be exactly one calendar day apart (compare pendingLastDay with entry.getKey()) before merging, so only true overnight splits are bridged.
```
- **nit** [correctness] line 314 — Merged route names now carry a non-ASCII en dash (U+2013); previously Timeline route names were pure ASCII dates, and some targets reachable via NavigationFormatConverter serialize with restricted charsets where the dash could be mangled. The writers are not visible in this diff, so I cannot name a concrete failing one.  _(correctness)_

---

